PR #50, opened 2017-08-24
Fixes #49, Use correct datetime format for ISO-8601

### LINKED ISSUE #49: Wrong Date format pattern
The date format pattern is wrong in above line!
The correct pattern should be "YYYY-MM-dd'T'HH:mm:ss.sss'Z'"

https://github.com/Adyen/adyen-java-api-library/blob/daea948e844affc48fa383529a30a42e9c5f1f51/src/main/java/com/adyen/serializer/DateTimeGMTSerializer.java#L34

lancergr:


rikterbeek:
@lancergr the added tests are failing can you have a look at that ?

src/main/java/com/adyen/serializer/DateTimeGMTSerializer.java
```diff
@@ -23,26 +23,22 @@
...
 
     /**
      * Format to ISO 8601 format: YYYY-MM-DDTHH:mm:ss.sssZ
-     *
-     * @param date
-     * @param typeOfSrc
-     * @param context
-     * @return
      */
+    @Override
```
rikterbeek: Why adding a override here ?

src/main/java/com/adyen/serializer/DateTimeGMTSerializer.java
```diff
@@ -23,26 +23,22 @@
...
     public JsonElement serialize(Date date, Type typeOfSrc, JsonSerializationContext context) {
         SimpleDateFormat formatter = new SimpleDateFormat(DATE_FORMAT);
+        formatter.setTimeZone(TimeZone.getTimeZone("GMT"));
```
rikterbeek: Is it possible to use the timezone of the merchant ?

src/main/java/com/adyen/serializer/DateTimeGMTSerializer.java
lancergr: adding override is a good practice to ensure that we override the correct method

src/main/java/com/adyen/serializer/DateTimeGMTSerializer.java
lancergr: since we add the Z in the end, we need to be in UTC
https://en.wikipedia.org/wiki/ISO_8601#UTC

PR #54, opened 2017-08-29
Make the Client stateless

### LINKED ISSUE #53: Make library Stateless
Nowadays library is stateful:
*HttpURLConnectionClient.createRequest* saves state at *httpConnection*.
This stateful causes a bad testing uses for API clients. If this class was stateless (minor changes), we can make a singleton wrapper with spring, improving the tests.

lancergr:
Fixes #53

### DIFF
--- a/src/main/java/com/adyen/Client.java
+++ b/src/main/java/com/adyen/Client.java
@@ -1,4 +1,4 @@
-/**
+/*
  *                       ######
  *                       ######
  * ############    ####( ######  #####. ######  ############   ############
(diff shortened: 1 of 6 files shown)


rikterbeek:
Hi @georgios,

Let's be backward compatible and keep the setHttpClient. It could be that customers want to use their own Client. we can do something like:

public ClientInterface createHttpClient() {
    if(this.httpClient != null) {
        try {
            return this.httpClient.getClass().newInstance();
        } catch (InstantiationException e) {
            e.printStackTrace();
        } catch (IllegalAccessException e) {
            e.printStackTrace();
        }
    }
    return new HttpURLConnectionClient();
}

public void setHttpClient(ClientInterface httpClient) {
    this.httpClient = httpClient;
}

PR #65, opened 2017-10-13
PW-298 Added container structures for MarketPay API

Aleffio:
Account, BankAccountDetail, ShareholderContact, AccountTransactionList, Transaction, AccountDetailBalance, Amount


src/main/java/com/adyen/model/AccountContainer.java
```diff
@@ -0,0 +1,83 @@
...
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("class AccountContainer {\n");
+
+        sb.append("    account: ").append(toIndentedString(account)).append("\n");
+        sb.append("}");
+        return sb.toString();
+    }
+
+    private String toIndentedString(Object o) {
```
epochcoder: Just randomly browsing here, but this method already exists in `Util.toIndentedString()`; it seems to be copied all over the model classes ;-)

rikterbeek: Hi @epochcoder,

Thanks for the comment. You are totally right. It is auto-generated code from swagger so that is why. Let's improve this separately from this PR because we need to update all our models. If you want you could create a PR to us with the fix if you need it right now otherwise we will fix it after finalizing this PR.

Regards,
Rik
Adyen

PR #83, opened 2018-01-26
Secured Fields helpers

lancergr:
**Description**
Added helpers for secured fields requests

**Tested scenarios**
Authorisation request with secured fields data


rikterbeek:
Hi @lancergr,

Please fix the code so it passes the tests.

Regards,
Rik
Adyen

rikterbeek:
We could expose a method setEncryptedCardToken that is setting the variables into the additionalData property as we have done for CSE called setCSEToken.

PR #86, opened 2018-02-23
Correction on return of exception (method authorise of Payment). 

davitwbd:
**Description**
The generic exception was changed to a specific exception called ApiException. This change was necessary because the generic exception return was hiding the returned error values.

**Tested scenarios**
Without this correction the exception of  authorise was returning just a com.adyen.service.exception.ApiException: HTTP Exception.

### DIFF
--- a/src/main/java/com/adyen/service/Payment.java
+++ b/src/main/java/com/adyen/service/Payment.java
@@ -48,8 +48,6 @@ public class Payment extends Service {
      * POST /authorise API call
      *
      * @param paymentRequest
-     * @return
-     * @throws Exception
      */
     public PaymentResult authorise(PaymentRequest paymentRequest) throws ApiException, IOException {
         String jsonRequest = GSON.toJson(paymentRequest);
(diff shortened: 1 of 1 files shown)


lancergr:
Hi @davitwbd 

Thanks for your Pull Request! The change looks fine because it makes the thrown exceptions more explicit. As long as the javadoc is updated, we can merge.

In the meantime I would advise to catch the ApiException separately.

Kind Regards,
Giorgos
Adyen

src/main/java/com/adyen/service/Payment.java
```diff
@@ -48,7 +51,7 @@ public Payment(Client client) {
      * @return
      * @throws Exception
```
lancergr: Could you please update the javadoc here?
It is also fine if you just remove "return" and "throws".

davitwbd: Ok! I will remove all ("return" and "throws").

lancergr:
Thanks @davitwbd

Issue #90, opened 2018-03-16
Util.implode bug


sanmibuh:
**Java version**: 1.7.0_80
**Library version**: 1.3.0
**Description**

We are using com.adyen.Util.Util#implode method for making the notification signing string.
But we found that if any element of the list is null, it adds `null` string instead of adding empty string.

lancergr:
Hi @sanmibuh 

Thanks for the Pull Request!

It would be interesting for me to know how did you bump into the problem of null Strings in the implode method.
The strings passed there should be escaped and this is done by the function https://github.com/Adyen/adyen-java-api-library/blob/develop/src/main/java/com/adyen/Util/HMACValidator.java#L76 which blanks null strings.

Regards,
Giorgos
Adyen

sanmibuh:
Hi @lancergr 
We are facing the problem when we are checking the notification signature.
But instead of making manually the signingData as described in the example 
https://docs.adyen.com/developers/payment-notifications/signature-for-notifications
we are reusing Util.implode directly.

`public boolean verify(final NotificationRequestItem notificationRequestItem)
      throws SignatureException {

        final List<String> signedDataList = new ArrayList<>(8);
        signedDataList.add(notificationRequestItem.getPspReference());
        signedDataList.add(notificationRequestItem.getOriginalReference());
        signedDataList.add(notificationRequestItem.getMerchantAccountCode());
        signedDataList.add(notificationRequestItem.getMerchantReference());
        final Amount amount = notificationRequestItem.getAmount();
        signedDataList.add(amount.getValue().toString());
        signedDataList.add(amount.getCurrency());
        signedDataList.add(notificationRequestItem.getEventCode());
        signedDataList.add(String.valueOf(notificationRequestItem.isSuccess()));

        /*
            NOTE Util.implode has a bug https://github.com/Adyen/adyen-java-api-library/issues/90
            Until fix, we need to make the signingData with other utility
         */
        //final String signingData = Util.implode(SIGNED_DATA_SEPARATOR, signedDataList);
        final String signingData = StringUtils.join(signedDataList, SIGNED_DATA_SEPARATOR);
         ....`

Apache lang3 StringUtils do the same thing, but we don't want to add another dependency only for this if the Adyen library has the utility for do it :)

Another question of this, is that it could be helpfully overloading methods on HMACValidator utility for notifications:

- calculateHMAC(NotificationRequestItem notificationRequestItem, NotificationRequestItem notificationRequestItem)
[...]

PR #91, opened 2018-03-16
[issue90]: Fix Util.implode.

sanmibuh:
**Description**
Now adds empty string when any element of the list is null.

**Tested scenarios**
Using utility for making the notification signing string.

**Fixed issue**:  #90


sanmibuh:
Updated the PR with the new utils for manage notification's HMAC signatures

lancergr:
Thanks @sanmibuh for the Pull Request!

Kind Regards,
Giorgos
Adyen

PR #99, opened 2018-06-26
Feature/checkout api pw 422

rkewlani:
Added/updated files for

1.API key support for user authentication.
2.Checkout and Checout Utility APIs.
Following APIs are added:
Checkout:
/paymentSession
 /payments/result
 /paymentMethods
 /payments
 /payments/details
Checkout Utility:
/originKeys

More test cases for API key are in progress. I will add them soon.


src/main/java/com/adyen/ApiKeyAuthenticatedService.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * <h1>ApiKeyAuthenticatedService</h1>
+ * The ApiKeyAuthenticatedService is an interface to enable a child service to support API key authentication feature.
+ */
+public class ApiKeyAuthenticatedService extends Service {
+
+    protected ApiKeyAuthenticatedService(Client client) {
+        super(client);
+        setApiKeySupported(true);
```
lancergr: let's better rename to setApiKeyRequired

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -43,11 +43,14 @@
...
      */
     @Override
-    public String request(String requestUrl, String requestBody, Config config) throws IOException, HTTPClientException {
+    public String request(String requestUrl, String requestBody, Config config, boolean isApiKeySupported) throws IOException, HTTPClientException {
         HttpURLConnection httpConnection = createRequest(requestUrl, config.getApplicationName());
+        if(isApiKeySupported)
+        {
+            setApiKey(httpConnection, config.getApiKey());
+        }
         setBasicAuthentication(httpConnection, config.getUsername(), config.getPassword());
```
lancergr: if api key is required, we shouldn't set basic auth headers

src/main/java/com/adyen/httpclient/ClientInterface.java
```diff
@@ -25,8 +25,10 @@
 import com.adyen.Config;
 
 public interface ClientInterface {
+
```
lancergr: "post" is used for DL and doesn't set any authentication headers. Why do we need to introduce postWithApiKeyFlag?

To be backwards compatible we can keep "request" method as one that uses basic auth and just introduce requestWithApiKey or smth for the one using x-api-key

src/main/java/com/adyen/model/checkout/Recurring.java
```diff
@@ -0,0 +1,241 @@
...
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+import java.util.Objects;
+
+/**
+ * Recurring
+ */
+public class Recurring {
```
lancergr: where do we use that class?

src/main/java/com/adyen/model/checkout/RecurringDetail.java
```diff
@@ -0,0 +1,263 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * RecurringDetail
+ */
+public class RecurringDetail {
```
lancergr: where do we use that class?

rkewlani: This class is an auto generated model used in oneClickPaymentMethods mthod of PaymentMethodsResponse class. Please see specification below. I believe oneClickPaymentMethods should be called by calling APIs (client APIs receiving the PaymentMethodsResponse ).
PaymentMethodsResponse:
      properties:
        oneClickPaymentMethods:
          description: Detailed list of one-click payment methods.
          items:
            $ref: '#/components/schemas/RecurringDetail'
          type: array

src/main/java/com/adyen/model/checkout/Recurring.java
rkewlani: This class is provided in specification but not used anywhere. For now, I am deleting it and will update content team to remove it from specification if not needed on their side too. thanks.

src/main/java/com/adyen/httpclient/ClientInterface.java
rkewlani: Thanks @lancergr 

Removed API authentication for post.

About request method backward compatibility, I believe this method is not externally exposed. This is part of internal client (HttpURLConnectionClient) used to send request to core system and request method is always called from Resource class. Please share your views more on this.

src/main/java/com/adyen/ApiKeyAuthenticatedService.java
rkewlani: done!

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
rkewlani: Basic authentication was set because core system checks username and password if API key is not present. Today discussed with core team and they confirmed that we should ignore username and password if API key is required. Hence incorporating the suggested changes and not setting basic auth if API key is required. thanks.

src/main/java/com/adyen/httpclient/ClientInterface.java
```diff
@@ -14,19 +14,20 @@
...
-import com.adyen.Config;
 
 public interface ClientInterface {
 
-	String request(String endpoint, String json, Config config) throws IOException, HTTPClientException;
+    String post(String endpoint, Map<String, String> postParameters, Config config) throws IOException, HTTPClientException;
 
-	String post(String endpoint, Map<String, String> postParameters, Config config) throws IOException, HTTPClientException;
+    String request(String endpoint, String json, Config config, boolean isApiKeyRequired) throws IOException, HTTPClientException;
 }
```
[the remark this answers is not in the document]
rkewlani: no. reverted back to original.

src/main/java/com/adyen/service/resource/checkout/Payments.java
```diff
@@ -0,0 +1,37 @@
...
+import com.adyen.service.Resource;
+
+import java.util.Arrays;
+
+public class Payments extends Resource {
+
+    public Payments(Service service) {
+        super(service, service.getClient().getConfig().getCheckoutEndpoint() + "/" + Client.CHECKOUT_API_VERSION + "/payments",
+                Arrays.asList("merchantAccount", "reference", "amount", "returnUrl", "paymentMethod"));
+
```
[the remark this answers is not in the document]
rkewlani: done!

src/main/java/com/adyen/service/CheckoutUtility.java
```diff
@@ -0,0 +1,61 @@
...
+import com.adyen.service.resource.checkoututility.OriginKeys;
+import com.google.gson.reflect.TypeToken;
+
+import java.io.IOException;
+
+public class CheckoutUtility extends ApiKeyAuthenticatedService {
+
+    private OriginKeys originKeys;
+
+
```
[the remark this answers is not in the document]
rkewlani: done!

src/main/java/com/adyen/service/CheckoutUtility.java
```diff
@@ -0,0 +1,61 @@
...
+    public CheckoutUtility(Client client) {
+
```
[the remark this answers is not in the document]
rkewlani: done!

src/main/java/com/adyen/service/CheckoutUtility.java
```diff
@@ -0,0 +1,61 @@
...
+        super(client);
+        originKeys = new OriginKeys(this);
+
```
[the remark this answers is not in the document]
rkewlani: done!

src/main/java/com/adyen/service/Resource.java
```diff
@@ -67,6 +67,7 @@ public String request(String json) throws ApiException, IOException {
             ApiException apiException = new ApiException(e.getMessage(), e.getCode());
             apiException.setError(apiError);
 
+
```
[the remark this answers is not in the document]
rkewlani: done!

src/main/java/com/adyen/service/Checkout.java
```diff
@@ -0,0 +1,132 @@
...
+        String jsonResult = paymentSession.request(jsonRequest);
+        return GSON.fromJson(jsonResult, new TypeToken<PaymentSessionResponse>() {
+        }.getType());
+    }
+
+    /**
+     * POST payments/result API call
+     *
+     * @param paymentResultRequest
+     * @return paymentVerificationResponse
```
lancergr: we better rename this doc-block return to PaymentResultResponse or just remove the param,return,throws since they don't add any additional info

rkewlani: done

src/main/java/com/adyen/service/Checkout.java
```diff
@@ -0,0 +1,132 @@
...
+        return GSON.fromJson(jsonResult, new TypeToken<PaymentsResponse>() {
+        }.getType());
+
+    }
+
+    /**
+     * POST /paymentSession API call
+     *
+     * @param paymentSessionRequest
+     * @return paymentSetupResponse
```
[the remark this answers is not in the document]
rkewlani: done

src/main/java/com/adyen/httpclient/ClientInterface.java
rkewlani: I checked more in documentations for this.  I found "Proxy configuration" in readme file and it is making object of HttpURLConnectionClient class. I think your point is good to keep old request method, just in case someone ever uses it in external code. Updated code for the same. thanks.

PR #106, opened 2018-07-25
make connectTimeout configurable for HttpURLConnection

NadirHelix:
**Description**
Referring to the support-ticket #318515 I added a simple change in the HttpURLConnectionClient to make the connectTimeout for HttpURLConnection configurable which can be important when using the proxy.

### DIFF
--- a/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
+++ b/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
@@ -38,6 +38,7 @@ import java.util.Scanner;
 public class HttpURLConnectionClient implements ClientInterface {
     private static final String CHARSET = "UTF-8";
     private Proxy proxy;
+    private int connectTimeout = -1;
 
     /**
      * Does a POST request.
@@ -122,6 +123,11 @@ public class HttpURLConnectionClient implements ClientInterface {
         } else {
             httpConnection = (HttpURLConnection) targetUrl.openConnection();
         }
+        
+        if (connectTimeout >= 0) {
+        	httpConnection.setConnectTimeout(connectTimeout);
+        }
+        
         httpConnection.setUseCaches(false);
         httpConnection.setDoOutput(true);
         httpConnection.setRequestMethod("POST");
@@ -200,5 +206,13 @@ public class HttpURLConnectionClient implements ClientInterface {
 
     public void setProxy(Proxy proxy) {
         this.proxy = proxy;
+    }
+    
+    public int getConnectTimeout() {
+    	return connectTimeout;
+    }
+    
+    public void setConnectTimeout(int connectTimeout) {
+    	this.connectTimeout = connectTimeout;
     }
 }
--- a/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
+++ b/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
@@ -38,7 +38,7 @@ import java.util.Scanner;
 public class HttpURLConnectionClient implements ClientInterface {
     private static final String CHARSET = "UTF-8";
     private Proxy proxy;
-    private int connectTimeout = -1;
+    private Integer connectionTimeoutMillis;
 
     /**
      * Does a POST request.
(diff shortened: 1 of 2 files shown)


src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -38,6 +38,7 @@
 public class HttpURLConnectionClient implements ClientInterface {
     private static final String CHARSET = "UTF-8";
     private Proxy proxy;
+    private int connectTimeout = -1;
```
rkewlani: We can give it a more meaningful name for example:  connectionTimeoutMillis

NadirHelix: Naming agreed. Could you explain a little bit more what you mean with Config and Environment?
Shall I add a field for timeout in Config as ell and check it in the HttpURLConnectionClient.java ?
If yes, which one should have priority?

rkewlani: Yes. A new field will be added in Config.java.
You may check an example of how Configuration setup for (x-api-key) is done here:
https://adyen.github.io/adyen-java-api-library/index.html

Usually, Config object is used for setting the environment/configuration settings (As shown in example) and HttpURLConnectionClient is not directly created so only Config object is setting connctionTimeOut.

Since you are following proxy settings section, hence you are creating a client directly.

So, in short, either Config or HttpURLConnectionClient will be used for setting timeout so at a time only one value will be there.

NadirHelix:
> 1.  Use of Integer in place of int is not recommended unless required. Unnecessary usage of Integer leads to performance drawbacks. Please change it to int. 

ok.

> 2. HttpURLConnectionClient class has changes for connectionTimeout field. Please refer to other fields like api-key and add similar validations and changes.

I'm not sure what you mean. Can you please provide an example.

> 3. Not sure why config.application is updated to config at some places. Can you please revert these changes if done by mistake ?

The changes were on purpose to avoid adding a 3rd argument. But if you think it's better to keep it separated on that level, I can adjust it.

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -38,6 +38,7 @@
 public class HttpURLConnectionClient implements ClientInterface {
     private static final String CHARSET = "UTF-8";
     private Proxy proxy;
+    private Integer connectionTimeoutMillis;
```
rkewlani: Please update this to int connectionTimeoutMillis;

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -54,7 +55,7 @@ public String request(String requestUrl, String requestBody, Config config) thro
 
     @Override
     public String request(String requestUrl, String requestBody, Config config, boolean isApiKeyRequired) throws IOException, HTTPClientException {
-        HttpURLConnection httpConnection = createRequest(requestUrl, config.getApplicationName());
+        HttpURLConnection httpConnection = createRequest(requestUrl, config);
```
rkewlani: Why this is changed?

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -122,16 +123,30 @@ private HttpURLConnection createRequest(String requestUrl, String applicationNam
...
+        
+        setConnectionTimeout(httpConnection, config);
+        
         httpConnection.setUseCaches(false);
         httpConnection.setDoOutput(true);
         httpConnection.setRequestMethod("POST");
 
         httpConnection.setRequestProperty("Accept-Charset", CHARSET);
-        httpConnection.setRequestProperty("User-Agent", String.format("%s %s%s", applicationName, Client.USER_AGENT_SUFFIX, Client.LIB_VERSION));
+        httpConnection.setRequestProperty("User-Agent", String.format("%s %s%s", config.getApplicationName(), Client.USER_AGENT_SUFFIX, Client.LIB_VERSION));
```
rkewlani: Same as above, why this is updated?

NadirHelix:
Ok I updated the PR according to your questions but I still do not understand the request regarding changes and validation. 
Please provide a code example so I can understand how it should look like.

rkewlani:
Now changes are looking better. Please refer to the field apiKey. We can make similar changes for connectionTimeOut.

NadirHelix:
@rkewlani could you please answer my last quastion?

rkewlani:
Hi Nadir,
I am happy to inform you that  Release 1.5.3 is published and now connection time out is configurable.
I am closing this pull request since this feature is available in latest release. 
Please let me know if you have any questions.

Regards,
Rajni

rkewlani:
This feature is released in 1.5.3.

PR #112, opened 2018-09-06
Added a new setter method to set PaymentDetailRequest properties.

rkewlani:


rkewlani:
new setter method to set PaymentDetailRequest properties.

PR #113, opened 2018-09-17
Feature/pw 622 connection timeout

rkewlani:
support to make ConnectionTimeOut configurable for Http client.

### DIFF
--- a/src/main/java/com/adyen/Client.java
+++ b/src/main/java/com/adyen/Client.java
@@ -63,30 +63,27 @@ public class Client {
         this.config.setApplicationName(applicationName);
     }
 
-    public Client(String username, String password, Environment environment, String applicationName, int connectionTimeoutMillis) {
+    public Client(String username, String password, Environment environment, int connectionTimeoutMillis) {
 
         this.config = new Config();
         this.config.setUsername(username);
         this.config.setPassword(password);
         this.setEnvironment(environment);
-        this.config.setApplicationName(applicationName);
         this.config.setConnectionTimeoutMillis(connectionTimeoutMillis);
     }
 
-    public Client(String apiKey, Environment environment, String applicationName) {
+    public Client(String apiKey, Environment environment) {
 
         this.config = new Config();
         this.config.setApiKey(apiKey);
         this.setEnvironment(environment);
-        this.config.setApplicationName(applicationName);
     }
 
-    public Client(String apiKey, Environment environment, String applicationName, int connectionTimeoutMillis) {
+    public Client(String apiKey, Environment environment, int connectionTimeoutMillis) {
 
         this.config = new Config();
         this.config.setApiKey(apiKey);
         this.setEnvironment(environment);
-        this.config.setApplicationName(applicationName);
         this.config.setConnectionTimeoutMillis(connectionTimeoutMillis);
     }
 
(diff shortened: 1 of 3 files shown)


rikterbeek:
Hi @rkewlani shall we add an overload as well on the constructor so you can easily set the timeout without creating your own config object ?

rkewlani:
> object

We have not defined constructor explicitly in HttpURLConnectionClient so overloading constructor means we will also need to explicitly provide default constructor.
Also, config object will be always needed because we are providing authentication and other required configuration information in Config object.

rikterbeek:
Hi @rkewlani Ment overloading of the Client object constructor so:
Client client = new Client("YourWSUser", "YourWSPassword", Environment.TEST, "myTestPayment", "TIMEOUT_IN_MS");

rkewlani:
> Hi @rkewlani Ment overloading of the Client object constructor so:
> Client client = new Client("YourWSUser", "YourWSPassword", Environment.TEST, "myTestPayment", "TIMEOUT_IN_MS");

Indeed will be helpful. I have overloaded constructor for apiKey as well. thanks @rikterbeek !

src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -56,12 +56,16 @@ public String request(String requestUrl, String requestBody, Config config) thro
...
         String apiKey = config.getApiKey();
+        int connectionTimeoutMillis = config.getConnectionTimeoutMillis();
         // Use Api key if required or if provided
         if (isApiKeyRequired || (apiKey != null && !apiKey.isEmpty())) {
             setApiKey(httpConnection, apiKey);
         } else {
             setBasicAuthentication(httpConnection, config.getUsername(), config.getPassword());
         }
+        if (connectionTimeoutMillis > 0) {
+            setConnectionTimeout(httpConnection, connectionTimeoutMillis);
```
lancergr: since the arguments and number of lines are the same, why not just use the httpConnection.setConnectTimeout(connectionTimeout) instead of wrapping it within a helper function?

rkewlani: It will give more flexibility to set it directly for HttpURLConnectionClient object even after object is already created and later connectiontimeout needs to be set. If we directly use httpConnection.setConnectionTimout(), we will loose this flexibility.

src/main/java/com/adyen/Client.java
```diff
@@ -63,6 +63,33 @@ public Client(String username, String password, Environment environment, String
...
+
+    public Client(String apiKey, Environment environment, String applicationName, int connectionTimeoutMillis) {
+
+        this.config = new Config();
+        this.config.setApiKey(apiKey);
+        this.setEnvironment(environment);
+        this.config.setApplicationName(applicationName);
+        this.config.setConnectionTimeoutMillis(connectionTimeoutMillis);
+    }
+
```
rikterbeek: let's add overloads as well for ApplicationName as this should not be a mandatory field

rkewlani: It is not mandatory. We can pass null value if we don't want to set applicationName. Point here is- we should not overload constructors for all possible combinations if we can pass null for some optional parameters.

rikterbeek: Understand, when we add applicationInfo we can remove this field so let's add only constructors without applicationName for now.

PR #120, opened 2018-10-01
PW-647: Fix Java library for checkout dynamic urls

Aleffio:


src/main/java/com/adyen/service/CheckoutUtility.java
```diff
@@ -36,6 +36,10 @@
 
     public CheckoutUtility(Client client) {
         super(client);
+        if (client.getConfig().getCheckoutEndpoint() == null || client.getConfig().getCheckoutEndpoint().isEmpty()) {
+            String message = "Please specify the unique identifier when setting the environment";
+            throw new IllegalArgumentException(message);
+        }
```
rikterbeek: duplicate code, let's find a better spot to reduct duplication

src/main/java/com/adyen/Client.java
```diff
@@ -87,23 +89,52 @@ public Client(String apiKey, Environment environment, int connectionTimeoutMilli
...
             this.config.setHppEndpoint(HPP_LIVE);
-            this.config.setCheckoutEndpoint(CHECKOUT_ENDPOINT_LIVE);
-        } else {
-            // throw exception
+            if (liveEndpointUrlPrefix != null && ! liveEndpointUrlPrefix.isEmpty()) {
+                this.config.setEndpoint(ENDPOINT_PROTOCOL + liveEndpointUrlPrefix + ENDPOINT_LIVE_SUFFIX);
+                this.config.setCheckoutEndpoint(ENDPOINT_PROTOCOL + liveEndpointUrlPrefix + CHECKOUT_ENDPOINT_LIVE_SUFFIX);
+            } else {
+                this.config.setEndpoint(ENDPOINT_LIVE);
+            }
```
rikterbeek: Maybe best to setCheckoutEndpoint to null so if you first do a setEnvironment(TEST) and then setEnvironment(LIVE) you don't use the test endpoint and our exception can be triggered. Now it will ignore it as it has a value.

PR #124, opened 2018-10-03
Release 1.5.4

lancergr:


rikterbeek:
@lancergr let's only do a new release where we fixed checkout endpoint.

PR #133, opened 2018-10-26
Added OSGI manifest

lancergr:
Added OSGI manifest
Added checkstyles
Changed Fund.AccountHolderBalance to Fund.accountHolderBalance


src/main/java/com/adyen/service/Modification.java
```diff
@@ -1,19 +1,19 @@
 /**
```
Aleffio: please fix the copyright

PR #137, opened 2018-10-29
Added paymentMethod DTO for /payments call

lancergr:
**Description**
Added paymentMethod DTO for /payments call

**Tested scenarios**
/payments with CC and APM


src/main/java/com/adyen/model/checkout/PaymentsRequest.java
```diff
@@ -639,26 +615,26 @@ public PaymentsRequest addEncryptedCardData(String encryptedCardNumber, String e
...
         return this;
     }
 
     public PaymentsRequest addOneClickData(String recurringDetailReference, String encryptedSecurityCode) {
-        this.paymentMethod = new HashMap<>();
-        this.paymentMethod.put(METHOD_TYPE, TYPE_SCHEME);
-        this.paymentMethod.put(RECURRING_DETAIL_REFERENCE, recurringDetailReference);
-        this.paymentMethod.put(ENCRYPTED_SECURITY_CODE, encryptedSecurityCode);
+        DefaultPaymentMethodDetails paymentMethodDetails = new DefaultPaymentMethodDetails();
+        this.paymentMethod = paymentMethodDetails;
```
rkewlani: minor: exchange of 634 and 637 will make it more readable.

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
```diff
@@ -593,41 +581,29 @@ public void setOrderReference(String orderReference) {
...
     }
 
     public PaymentsRequest addEncryptedCardData(String encryptedCardNumber, String encryptedExpiryMonth, String encryptedExpiryYear, String encryptedSecurityCode, String holderName) {
-        this.paymentMethod = new HashMap<>();
-        this.paymentMethod.put(METHOD_TYPE, TYPE_SCHEME);
-        this.paymentMethod.put(ENCRYPTED_CARD_NUMBER, encryptedCardNumber);
-        this.paymentMethod.put(ENCRYPTED_EXPIRY_MONTH, encryptedExpiryMonth);
-        this.paymentMethod.put(ENCRYPTED_EXPIRY_YEAR, encryptedExpiryYear);
+        DefaultPaymentMethodDetails paymentMethodDetails = new DefaultPaymentMethodDetails();
+        this.paymentMethod = paymentMethodDetails;
```
rkewlani: minor: exchange of 599 and 601 will make it more readable.

src/test/java/com/adyen/CheckoutTest.java
```diff
@@ -196,6 +204,25 @@ public void TestPaymentMethodsFailureMissingIdentifierOnLive() throws Exception
 
     }
 
+    @Test
```
rkewlani: I think it will be a good idea to add one more test case for DefaultPaymentDetails class coverage.

lancergr: The intention here is to test the inject-ability of custom method details classes that implement PaymentMethodDetails. Generally we don't test model POJO classes and DefaultPaymentDetails is one of them, but feel free to send a suggestion of the extra test case if needed

lancergr: Testing PaymentMethodDetails is done at L211 and inject-ability of custom details class on L215.
Do you suggest to add one more test case or to remove the test case of L215 ? Could you give example?

rkewlani: If it is already covered in L211 then we are good to go :)

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
lancergr: done

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
lancergr: done

PR #140, opened 2018-11-02
Custom read timeout for HttpURLConnection

marinan87:
**Description**
See Adyen Support ticket #447790

**Tested scenarios**
Tested locally simulating network outage after establishing initial connection.

**Fixed issue**:

### DIFF
--- a/src/main/java/com/adyen/Client.java
+++ b/src/main/java/com/adyen/Client.java
@@ -69,6 +69,26 @@ public class Client {
         this.config.setApplicationName(applicationName);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


lancergr:
Thank you @marinan87 so much for opening a Pull Request for that!

I believe removing client constructors is good idea for keeping simplicity. Could you please revert the removed client constructors and mark them as deprecated so that the backwards compatibility is maintained and we can remove them in a new major version?

Best Regards,
Giorgos
Adyen

src/main/java/com/adyen/Client.java
```diff
@@ -176,4 +183,9 @@ public void setApplicationName(String applicationName) {
         this.config.setApplicationName(applicationName);
     }
 
+    public void setTimeouts(int connectionTimeoutMillis, int readTimeoutMillis) {
```
rkewlani: Setter methods should be separate for connectiontimeout and readtimeout.

Issue #157, opened 2019-02-27
Possible NullPointerException on PaymentResultResponse & PaymentRequest


sanmibuh:
**Java version**: 1.8
**Library version**: 2.0.0
**Description**
**PaymentResultResponse** and **PaymentRequest** has a possible NullPointerException if we call to 
_putAdditionalDataItem_ or _putMetadataItem_ without calling first to _additionalData_ or _metadata_

rkewlani:
Hi @sanmibuh,
Thanks for PR to fix this. We will merge and release this once review is done.

Thanks,
Rajni

PR #158, opened 2019-02-27
Fix npe

sanmibuh:
**Description**
**PaymentResultResponse** and **PaymentRequest** has a possible NullPointerException if we call to 
_putAdditionalDataItem_ or _putMetadataItem_ without calling first to _additionalData_ or _metadata_
The way to solve the issue is the same as applied on **PaymentsResponse**

Plus: Removed on those 2 classes the use of StringBuilder on **toString** 

**Tested scenarios**
NullPointerException if we call to 
_putAdditionalDataItem_ or _putMetadataItem_ without calling first to _additionalData_ or _metadata_

**Fixed issue**:  #157

### DIFF
--- a/src/main/java/com/adyen/model/checkout/PaymentResultResponse.java
+++ b/src/main/java/com/adyen/model/checkout/PaymentResultResponse.java
@@ -238,19 +238,16 @@ public class PaymentResultResponse {
 
     @Override
     public String toString() {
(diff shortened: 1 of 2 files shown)


rkewlani:
Hi @sanmibuh,
Thanks for PR. We will merge and release this once review is done.

Thanks,
Rajni

PR #162, opened 2019-03-11
[PW-1090]: Added response codes for 3DS 2.0

rkewlani:
**Description**

**Tested scenarios**

**Fixed issue**:


src/main/java/com/adyen/model/checkout/PaymentsResponse.java
```diff
@@ -410,6 +410,8 @@ private String toIndentedString(Object o) {
         RECEIVED("Received"),
         REDIRECTSHOPPER("RedirectShopper"),
         PRESENTTOSHOPPER("PresentToShopper"),
+        IDENTIGYSHOPPER("IdentifyShopper"),
```
Aleffio:
```suggestion
        IDENTIFYSHOPPER("IdentifyShopper"),
```

PR #170, opened 2019-03-29
Add sepa support

AlexandrosMor:


src/test/java/com/adyen/CheckoutTest.java
```diff
@@ -253,6 +257,35 @@ public void TestPaymentMethodDetails() {
...
 
+    @Test
+    public void TestSepaPaymentMethodDetails() {
+        DefaultPaymentMethodDetails defaultPaymentMethodDetails=new DefaultPaymentMethodDetails();
+        defaultPaymentMethodDetails.type(TYPE_SEPA_DIRECT_DEBIT);
+        defaultPaymentMethodDetails.setSepaOwnerName("A. Schneider");
+        defaultPaymentMethodDetails.setSepaIbanNumber("DE87123456781234567890");
+
+        PaymentsRequest paymentsRequest = createPaymentsCheckoutRequest();
+        paymentsRequest.setApplicationInfo(null);
```
rkewlani: This is redundant. please remove null assignment.

src/test/java/com/adyen/CheckoutTest.java
```diff
@@ -253,6 +257,35 @@ public void TestPaymentMethodDetails() {
                              + "}", jsonRequest);
     }
 
+    @Test
+    public void TestSepaPaymentMethodDetails() {
+        DefaultPaymentMethodDetails defaultPaymentMethodDetails=new DefaultPaymentMethodDetails();
+        defaultPaymentMethodDetails.type(TYPE_SEPA_DIRECT_DEBIT);
+        defaultPaymentMethodDetails.setSepaOwnerName("A. Schneider");
```
rkewlani: This type constant is not needed.

src/test/java/com/adyen/CheckoutTest.java
AlexandrosMor: We need to remove the application info because the version is always updating and the mock request is not.

PR #177, opened 2019-04-10
[PW-828] Add application info to adyen-java-api-library

martinsrenato:
**Description**
ApplicationInfo sets library values on constructor and `setAdyenLibrary` removed.
ApplicationInfo added to every request, with prefilled library name + version.
Added request tests.

**Tested scenarios**
- Adding applicationInfo into request does not override the library name + version
- Not adding applicationInfo into request keeps default applicationInfo with library name + version request


src/main/java/com/adyen/model/AbstractPaymentRequest.java
```diff
@@ -758,6 +756,9 @@ public ApplicationInfo getApplicationInfo() {
     }
 
     public void setApplicationInfo(ApplicationInfo applicationInfo) {
+        if (applicationInfo == null) {
```
martinsrenato: Another option would be to set `this.applicationInfo = new ApplicationInfo()`

src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
```diff
@@ -102,6 +106,9 @@ public ApplicationInfo getApplicationInfo() {
     }
 
     public void setApplicationInfo(ApplicationInfo applicationInfo) {
+        if (applicationInfo == null) {
```
KadoBOT: Maybe this can me externalised somehow? I see that this piece of code is repeated, which makes future changes a bit harder.

martinsrenato: It's already externalised for some requests on `AbstractPaymentRequest ` and `AbstractModificationRequest` (those are extended on other classes), but I'll check if we can to the same for all of them

src/main/java/com/adyen/model/checkout/PaymentSessionRequest.java
```diff
@@ -538,6 +534,9 @@ public ApplicationInfo getApplicationInfo() {
     }
 
     public void setApplicationInfo(ApplicationInfo applicationInfo) {
+        if (applicationInfo == null) {
```
AlexandrosMor: Should we throw an exception if it is not set? It is not part of required fields

src/main/java/com/adyen/model/checkout/PaymentSessionRequest.java
```diff
@@ -538,6 +534,9 @@ public ApplicationInfo getApplicationInfo() {
...
+            throw new IllegalArgumentException("ApplicationInfo cannot be null");
```
AlexandrosMor: Should we throw an exception if it is not set? It is not part of required fields

src/main/java/com/adyen/model/AbstractPaymentRequest.java
```diff
@@ -758,6 +756,9 @@ public ApplicationInfo getApplicationInfo() {
...
+            throw new IllegalArgumentException("ApplicationInfo cannot be null");
+        }
```
rkewlani: We should not throw exception here.

src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
```diff
@@ -81,6 +81,10 @@
     @SerializedName("trustedShopper")
     private Boolean trustedShopper = null;
 
+    public PaymentMethodsRequest() {
+        applicationInfo = new ApplicationInfo();
+    }
+
```
rkewlani: This is not needed here. ApplicationInfo is used only in Payment submission.

src/test/java/com/adyen/PaymentMethodsRequestTest.java
```diff
@@ -0,0 +1,65 @@
+/**
```
rkewlani: This class not needed since we will not use ApplicationInfo in PaymentMethodRequest class.

src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
martinsrenato: We decided to remove the setter, so this is not an issue anymore.

src/main/java/com/adyen/model/checkout/PaymentSessionRequest.java
martinsrenato: We decided to remove the setter, so this is not an issue anymore.

src/main/java/com/adyen/model/AbstractPaymentRequest.java
martinsrenato: We decided to remove the setter, so this is not an issue anymore.

src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
martinsrenato: Reverted.

src/test/java/com/adyen/PaymentMethodsRequestTest.java
martinsrenato: Reverted.

src/test/java/com/adyen/ModificationRequestTest.java
```diff
@@ -0,0 +1,60 @@
+/**
```
Aleffio: Can you please apply the correct copyright to the new files?

martinsrenato: Done

PR #187, opened 2019-05-02
[PW-883] Nexo Model (Terminal API)

martinsrenato:
**Description**
Model generated from the Nexo xsd file


tjerkw:
@martinsrenato Could you mention which command you used to generate the model files from the xsd? (And could you post the xsd). 

Want todo the same for: https://github.com/tjerkw/adyen-kotlin-multiplatform-api-library

Could also do it manually, but this sounds easier.

PR #191, opened 2019-05-10
added models, services and mocking test for binlookup apis

rikterbeek:
**Description**
Added BinLookup API /getCostEstimate and /get3dsAvailability

**Tested scenarios**
Success and fail cases with different request data

### DIFF
--- a/src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
+++ b/src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
@@ -9,7 +9,7 @@ public class Get3dsAvailability extends Resource {
 
     public Get3dsAvailability(Service service) {
         super(service,
-                service.getClient().getConfig().getEndpoint() + "/pal/servlet/BinLookup/" + Client.BIN_LOOKUP_API_VERSION + "/get3dsAvailability",
+                service.getClient().getConfig().getEndpoint() + Client.BIN_LOOKUP_PAL_SUFFIX + Client.BIN_LOOKUP_API_VERSION + "/get3dsAvailability",
               Arrays.asList("merchantAccount", "cardNumber"));
     }
 }
--- a/src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
+++ b/src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
@@ -1,0 +1,14 @@
+package com.adyen.service.resource.binlookup;
+
+import java.util.Arrays;
+import com.adyen.Client;
+import com.adyen.Service;
+import com.adyen.service.Resource;
+
+public class Get3dsAvailability extends Resource {
+
+    public Get3dsAvailability(Service service) {
+        super(service, service.getClient().getConfig().getCheckoutEndpoint() + "/" + Client.BIN_LOOKUP_API_VERSION + "/get3dsAvailability",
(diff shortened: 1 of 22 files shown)


src/test/java/com/adyen/BinLookupTest.java
```diff
@@ -0,0 +1,93 @@
...
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for
+ * /get3dsAvailability
+ * /getCostEstimate
+ */
+public class BinLookupTest extends BaseTest {
+
+
```
rkewlani: It would be good to have some negative test scenarios as well.

src/test/resources/mocks/binlookup/get3dsavailability-success.json
```diff
@@ -0,0 +1,16 @@
+{
```
rkewlani: Same as above, we can add some mock responses for invalid scenarios like invalid data etc.

src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Arrays;
+import com.adyen.Client;
+import com.adyen.Service;
+import com.adyen.service.Resource;
+
+public class Get3dsAvailability extends Resource {
+
+    public Get3dsAvailability(Service service) {
+        super(service,
+                service.getClient().getConfig().getEndpoint() + "/pal/servlet/BinLookup/" + Client.BIN_LOOKUP_API_VERSION + "/get3dsAvailability",
```
rkewlani: we can add this as a constant in Apiconstants "/pal/servlet/BinLookup/" so that if URL changes we don't need to updated in 2 classes.

src/main/java/com/adyen/service/resource/binlookup/GetCostEstimate.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Arrays;
+import com.adyen.Client;
+import com.adyen.Service;
+import com.adyen.service.Resource;
+
+public class GetCostEstimate extends Resource {
+
+    public GetCostEstimate(Service service) {
+        super(service,
+                service.getClient().getConfig().getEndpoint() + "/pal/servlet/BinLookup/" + Client.BIN_LOOKUP_API_VERSION + "/getCostEstimate",
```
rkewlani: same as above. constant for URL.

src/test/java/com/adyen/BinLookupTest.java
martinsrenato: Indeed. Just added them.

src/test/resources/mocks/binlookup/get3dsavailability-success.json
martinsrenato: Added

src/main/java/com/adyen/service/resource/binlookup/Get3dsAvailability.java
martinsrenato: Added constant in Client with the other urls

src/main/java/com/adyen/service/resource/binlookup/GetCostEstimate.java
martinsrenato: Added constant in Client with the other urls

PR #211, opened 2019-05-28
[PW-1279] Terminal API certificate validation

martinsrenato:


src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -232,41 +242,47 @@ private String doPostRequest(HttpURLConnection httpConnection, String requestBod
...
+            } catch (GeneralSecurityException | IOException e) {
+                throw new HTTPClientException("Error loading certificate from path", e);
+            }
 
             // Create terminal-trusting host name verifier
             HostnameVerifier terminalHostsValid = new HostnameVerifier() {
                 public boolean verify(String host, SSLSession session) {
-                    if (host.matches(".+\\..+\\.terminal\\.adyen\\.com")) {
+                    String url = host + ":" + session.getPeerPort();
+                    if (url.matches(TERMINAL_API_HOST_REGEX) || url.matches(TERMINAL_API_IP_REGEX)) {
```
rikterbeek: let's validate this as well with @AlexandrosMor I am not sure this is right.

martinsrenato: Made the change to skip this validation entirely, like we talked earlier @rikterbeek @AlexandrosMor

PR #212, opened 2019-05-29
Adding fields for v49 of Checkout and Payments API and v50 of BinLookup

martinsrenato:


src/main/java/com/adyen/Client.java
```diff
@@ -35,7 +35,7 @@
     public static final String HPP_LIVE = "https://live.adyen.com/hpp";
     public static final String MARKETPAY_ENDPOINT_TEST = "https://cal-test.adyen.com/cal/services";
     public static final String MARKETPAY_ENDPOINT_LIVE = "https://cal-live.adyen.com/cal/services";
-    public static final String API_VERSION = "v40";
+    public static final String API_VERSION = "v46";
```
rkewlani: Should we not migrate to 49 directly ?

martinsrenato: Yes, my mistake.

Issue #214, opened 2019-06-04
PaymentMethodDetails object includes deliveryAddress, billingAddress, and personalDetails attriutes


tddouglas:
**Library version**: 2.0.0
**Description**
In the DefaultPaymentMethodDetails.java file, deliveryAddress, billingAddress, and personalDetails are listed as attributes of the paymentMethod object. However, you can't make a /payments call with deliveryAddress or billingAddress nested inside of the paymentMethod object (they aren't parsed by Adyen). Additionally, I don't know what the personalDetails field is, but it is not mentioned anywhere in the documentation or in our API explorer.

tddouglas:
https://github.com/Adyen/adyen-java-api-library/blob/74a3e82ad726b66246d950b67fed7ffbc72d0afd/src/main/java/com/adyen/model/checkout/DefaultPaymentMethodDetails.java#L28

martinsrenato:
Hi @jvalecillos !

You fix was included in the release version 2.5.0 of the library. Thanks for the input!

Regards,
Renato
Adyen

PR #215, opened 2019-06-06
Removing unused fields from payment request payload

jvalecillos:
**Description**
In conversation with support it was indicated that the object fields **personalDetails**, **billingAddress** and **deliveryAddress** are not used, or at least redundant considering that the content is already present in the root payload for [/payments](https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v46/payments) request.

I've checked the documentation, the API Explorer and the code itself and it seems that the fields are not referenced at all outside this code. Nevertheless, the answer about if the fields are needed it is up to the discretion of the maintainers.

**Tested scenarios**
The fields being removed were not originally referenced in the tests. Thus, this change could actually increase the coverage as a side effect of deleting unused code.

**Fixed issue**: #214


cyattilakiss:
Hi @jvalecillos ,

The personalDetails is unfortunately can be sent in the root or in the paymentMethod but indeed the deliveryAddress and billingAddress can be removed from the paymentMethod.

Best,
Attila
Adyen

martinsrenato:
Hi @jvalecillos!

I understand that the PersonalDetails object seems redundant considering the whole payload, but I checked with the team here at Adyen, and the PersonalDetails data on PaymentMethod is needed for some specific types of payments on the Checkout API platform.
On the other hand, the deliveryAddress and billingAddress can be removed, like @cyattilakiss explained before.

Do you think you can revert the changes to PersonalDetails on this PR? Than we can review and add the changes to a future release.

Thanks!

Renato
Adyen

jvalecillos:
Thanks for the clarification @cyattilakiss and @martinsrenato

As you indicated I have restored back the `personalDetails` in my PR. Now, only `deliveryAddress` and `billingAddress` will be removed.

martinsrenato:
Hey @jvalecillos!
Really sorry for taking so long to merge your PR. 
As you can see, the travis build is failing here. We had the same problem with other PRs on this project and after some investigation we discovered it was a config error.
We already fixed it on the develop branch and it's been working fine since. 
Would you mind merging the changes that we made in develop to this branch so the build can work again? Sorry to bother you with that.

Thanks!
Renato

jvalecillos:
Rebased from upstream `develop` branch.

PR #220, opened 2019-06-17
 Fix terminal abort request

martinsrenato:
PW-1384 - Terminal API AbortRequest results in exception


rkewlani:
Can you please check why build is failing ?

src/main/java/com/adyen/service/TerminalLocalAPI.java
```diff
@@ -65,6 +65,10 @@ public TerminalAPIResponse request(TerminalAPIRequest terminalAPIRequest, Securi
 
         String jsonResponse = localRequest.request(jsonEncryptedRequest);
 
+        if (jsonResponse == null || jsonResponse.isEmpty()) {
```
rkewlani: do we need to check "ok" here ?

AlexandrosMor: "Ok" is not a SaleToPoiResponse. I believe the response should be null

martinsrenato: Like @AlexandrosMor , for the local API in this case we only get null or a SaleToPoiResponse object as response, so I don't think there's a need to put "ok" here

PR #224, opened 2019-07-03
Fix ForceEntryMode fields / Instalment and CheckData field names

martinsrenato:

### DIFF
--- a/src/main/java/com/adyen/model/nexo/Instalment.java
+++ b/src/main/java/com/adyen/model/nexo/Instalment.java
@@ -50,7 +50,7 @@ public class Instalment {
      * The Instalment.
      */
     @XmlList
-    @XmlElement(name = "Instalment")
+    @XmlElement(name = "InstalmentType")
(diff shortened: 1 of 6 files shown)


src/main/java/com/adyen/model/nexo/Instalment.java
```diff
@@ -50,7 +50,7 @@
      * The Instalment.
      */
     @XmlList
-    @XmlElement(name = "Instalment")
+    @XmlElement(name = "InstalmentType")
```
rikterbeek: Should this not be just Instalment ?

martinsrenato: InstalmentType is the correct name according to the nexo specification.
It ended up wrong when I edited the xsd file to remove the Type suffix from the model names, but ended up renaming this field name.
This and CheckData.TypeCode (also fixed on this PR) where the only ones affected, I double checked it comparing to the original xsd.

PR #226, opened 2019-07-12
[PW-1398] Fix AuthenticationMethodType.OnlinePIN

martinsrenato:
Handling both "On**L**inePIN", "On**l**inePIN"


rikterbeek:
@martinsrenato add youtrack number

PR #232, opened 2019-08-10
Missing attrs added to toString method

afilardo:
We weren't logging this attrs, so I realized that they were missing.
I'm not sure if in the non-encrypted case, this is intentional, so I have left it as it was

**Description**

**Tested scenarios**

**Fixed issue**:

PR #236, opened 2019-08-21
Missing attrs added to toString method

martinsrenato:
**Description**
Original PR #232 
Original Commit https://github.com/Adyen/adyen-java-api-library/pull/232/commits/4c7659f1b8a68e0e7d292722f30290c7101b3de7
Reopened it to rebase code from develop.

> We weren't logging this attrs, so I realized that they were missing.
> I'm not sure if in the non-encrypted case, this is intentional, so I have left it as it was


rikterbeek:
@martinsrenato why do we need this into the string ? As it is still encrypted data better not to show + logging it. So I rather don't add this add all. What are the reasons of adding this ?

rikterbeek:
@rkewlani @martinsrenato indeed because you don't want to log encrypted card data.

PR #245, opened 2019-09-09
Throw specific exception instead of a generic one

H4rryp0tt3r:
**Description**
`authorise3D` method in `Payment.java` which is used by library consumers is throwing `Exception`, instead of throwing `ApiException`, `IOException`.  Hence consumers are forced to handle this generic exception.

**Tested scenarios**
As this is a small refactoring, no tests were needed.

**Fixed issue**:  
This is not an issue, and just a small refactoring.

### DIFF
--- a/src/main/java/com/adyen/service/Payment.java
+++ b/src/main/java/com/adyen/service/Payment.java
@@ -79,9 +79,10 @@ public class Payment extends Service {
      *
      * @param paymentRequest3d authorise3D
      * @return PaymentResult
-     * @throws Exception Exception
+     * @throws ApiException ApiException
+     * @throws IOException IOException
      */
-    public PaymentResult authorise3D(PaymentRequest3d paymentRequest3d) throws Exception {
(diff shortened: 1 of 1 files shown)


src/main/java/com/adyen/service/Payment.java
```diff
@@ -99,7 +100,7 @@ public PaymentResult authorise3D(PaymentRequest3d paymentRequest3d) throws Excep
      * @return PaymentResult
      * @throws Exception Exception
```
martinsrenato: Hello @H4rryp0tt3r !

Could you also fix the java code for `@throws` in this method? It seems our travis build is failing because javadocs style.

Regards,
Renato
Adyen

H4rryp0tt3r: Hello @martinsrenato!

I've added those changes and force pushed as it made sense to keep it in the same commit. However travis didn't pick those.

Thanks,
Nagesh

rkewlani:
Thanks for PR H4rryp0tt3r.
I am merging your PR.

PR #254, opened 2019-10-10
[PW-1649] Add SaleToAcquirerDataModel with application info prefilled applicati…

AlexandrosMor:
…on info

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/src/main/java/com/adyen/Util/Util.java
+++ b/src/main/java/com/adyen/Util/Util.java
@@ -20,16 +20,18 @@
  */
 package com.adyen.Util;
 
-import com.adyen.model.Amount;
+import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.Base64;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
+import com.adyen.model.Amount;
 
 public final class Util {
     private Util() {
@@ -154,5 +156,21 @@ public final class Util {
         fmt.setTimeZone(TimeZone.getTimeZone("GMT"));
 
         return fmt.format(sessionDate);
+    }
+
+    /**
+     * Encode the given string to base 64 string
+     * @param o string
+     * @return base 64 encoded string    "
+     */
+    public static String toBase64Encode(String input) {
+        if (input != null) {
+            try {
+                return Base64.getEncoder().encodeToString(input.getBytes("utf-8"));
+            } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
+            }
+        }
+        return null;
     }
 }
--- a/src/main/java/com/adyen/Util/Util.java
+++ b/src/main/java/com/adyen/Util/Util.java
@@ -156,16 +156,4 @@ public final class Util {
 
         return fmt.format(sessionDate);
     }
-
-    /**
-     * Encode the given string to base 64 string
-     * @param input string
-     * @return base 64 encoded string    "
-     */
-    public static String toBase64Encode(String input) {
-        if (input != null) {
-            return new String(Base64.encodeBase64(input.getBytes()));
-        }
-        return null;
-    }
 }
--- a/src/main/java/com/adyen/Util/Util.java
+++ b/src/main/java/com/adyen/Util/Util.java
@@ -20,17 +20,16 @@
  */
 package com.adyen.Util;
 
-import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
-import java.util.Base64;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
+import org.apache.commons.codec.binary.Base64;
 import com.adyen.model.Amount;
 
 public final class Util {
@@ -165,11 +164,7 @@ public final class Util {
      */
     public static String toBase64Encode(String input) {
         if (input != null) {
-            try {
-                return Base64.getEncoder().encodeToString(input.getBytes("utf-8"));
-            } catch (UnsupportedEncodingException e) {
-                e.printStackTrace();
-            }
+            return new String(Base64.encodeBase64(input.getBytes()));
         }
         return null;
     }
--- a/src/main/java/com/adyen/Util/Util.java
+++ b/src/main/java/com/adyen/Util/Util.java
@@ -160,7 +160,7 @@ public final class Util {
 
     /**
      * Encode the given string to base 64 string
-     * @param o string
+     * @param input string
      * @return base 64 encoded string    "
      */
     public static String toBase64Encode(String input) {
--- a/src/main/java/com/adyen/Util/Util.java
+++ b/src/main/java/com/adyen/Util/Util.java
@@ -29,7 +29,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.TimeZone;
-import org.apache.commons.codec.binary.Base64;
 import com.adyen.model.Amount;
 
 public final class Util {
--- a/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
+++ b/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
@@ -1,7 +1,5 @@
 package com.adyen.model.terminal;
 
-import java.io.UnsupportedEncodingException;
-import java.util.Base64;
 import java.util.Map;
 import java.util.Objects;
 import com.adyen.model.applicationinfo.ApplicationInfo;
--- a/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
+++ b/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
@@ -83,6 +83,6 @@ public class SaleToAcquirerDataModel {
 
     @Override
     public int hashCode() {
-        return Objects.hash(applicationInfo, additionalData,keyValueData,metaData,tenderOption);
+        return Objects.hash(applicationInfo, additionalData, keyValueData, metaData, tenderOption);
     }
 }
--- a/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
+++ b/src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
@@ -73,12 +73,16 @@ public class SaleToAcquirerDataModel {
         if (o == null || getClass() != o.getClass()) {
             return false;
         }
-        SaleToAcquirerDataModel that = (SaleToAcquirerDataModel) o;
-        return additionalData.equals(that.additionalData) && applicationInfo.equals(that.applicationInfo);
+        SaleToAcquirerDataModel saleToAcquirerDataModel = (SaleToAcquirerDataModel) o;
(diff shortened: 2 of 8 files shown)


src/test/java/com/adyen/ApplicationInfoTest.java
```diff
@@ -28,4 +28,20 @@ public void TestApplicationInfoPrefilledLibraryFields() {
...
+
+        SaleToAcquirerDataModel saleToAcquirerDataModel = new SaleToAcquirerDataModel();
+        String applicationInfoExpected = "{\n"
+                + "  \"applicationInfo\": {\n"
+                + "    \"adyenLibrary\": {\n"
+                + "      \"name\": \""+LIB_NAME+"\",\n"
+                + "      \"version\": \""+LIB_VERSION+"\"\n"
+                + "    }\n"
+                + "  }\n"
+                + "}";
```
msilvagarcia: I would prefer something that's clearer than a concatenation. Could be either an external JSON file or  converting a plain dictionary into JSON?

src/main/java/com/adyen/Util/Util.java
```diff
@@ -155,4 +156,16 @@ public static String calculateSessionValidity() {
...
         return fmt.format(sessionDate);
     }
+
+    /**
+     * Encode the given string to base 64 string
+     * @param input string
+     * @return base 64 encoded string    "
+     */
+    public static String toBase64Encode(String input) {
+        if (input != null) {
```
rikterbeek: should we not just do it. If it fails the exception is fine.

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public void setTenderOption(String tenderOption) {
+        this.tenderOption = tenderOption;
+    }
+
+    public String toJson() {
+        return PRETTY_PRINT_GSON.toJson(this);
+    }
+
```
rikterbeek: would be nice to add a to toBase64() here that does the trick.

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -83,4 +83,24 @@ public TerminalAPIResponse sync(TerminalAPIRequest terminalAPIRequest) throws IO
...
+     */
+    private static TerminalAPIRequest checkSaleToAcquirerDataApplicationInfo(TerminalAPIRequest terminalAPIRequest) {
+        if (terminalAPIRequest != null
+                && terminalAPIRequest.getSaleToPOIRequest() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData().getSaleToAcquirerData() == null) {
+            SaleToAcquirerDataModel saleToAcquirerDataModel = new SaleToAcquirerDataModel();
+            String saleToAcquirerDataJson = saleToAcquirerDataModel.toJson();
+            String saleToAcqurerDataJsonBase64 = Util.toBase64Encode(saleToAcquirerDataJson);
```
rikterbeek: would be nice to have that toBase64 method iplemented then you can use terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData().setSaleToAcquirerData(saleToAcquirerDataModel.toBase64());

src/test/java/com/adyen/ApplicationInfoTest.java
rikterbeek: this is to keep it readable. We could do map to json or something.

src/main/java/com/adyen/Util/Util.java
AlexandrosMor: The method is moved to SaleToAcquirerData class

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
AlexandrosMor: toBase64 method is added

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
```diff
@@ -0,0 +1,84 @@
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        SaleToAcquirerDataModel that = (SaleToAcquirerDataModel) o;
+        return additionalData.equals(that.additionalData) && applicationInfo.equals(that.applicationInfo);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(applicationInfo, additionalData);
```
rkewlani: other fields are missing to calculate hash

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -56,8 +56,8 @@ public TerminalCloudAPI(Client client) {
      * @throws ApiException ApiException
      */
     public String async(TerminalAPIRequest terminalAPIRequest) throws IOException, ApiException {
+        terminalAPIRequest = this.checkSaleToAcquirerDataApplicationInfo(terminalAPIRequest);
```
rkewlani: this should be get instead of check.

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -83,4 +82,22 @@ public TerminalAPIResponse sync(TerminalAPIRequest terminalAPIRequest) throws IO
...
         return terminalAPIResponse;
     }
+
+    /**
+     * Checks if SaleToAcquirerData is set. If it is not creates an encoded base 64 string with application info.
+     *
+     * @param terminalAPIRequest TerminalAPIRequest
+     * @return TerminalAPIRequest
+     */
+    private static TerminalAPIRequest getApplicationInfo(TerminalAPIRequest terminalAPIRequest) {
```
rkewlani: method name is not appropriate according to operation. it is setting saletoacquiredata and not getting applicationinfo.

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -83,4 +82,22 @@ public TerminalAPIResponse sync(TerminalAPIRequest terminalAPIRequest) throws IO
...
+        if (terminalAPIRequest != null
+                && terminalAPIRequest.getSaleToPOIRequest() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData() != null
+                && terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData().getSaleToAcquirerData() == null) {
```
msilvagarcia: This validation makes more sense in the `TerminalAPIRequest` object as a `isSaleToAcquirerDataSet` method

src/main/java/com/adyen/service/PosPayment.java
```diff
@@ -48,7 +48,7 @@ public PosPayment(Client client) {
      * @throws ApiException ApiException
      */
     public ConnectedTerminalsResponse connectedTerminals(ConnectedTerminalsRequest connectedTerminalsRequest) throws IOException, ApiException {
-        String jsonRequest = GSON.toJson(connectedTerminalsRequest);
+       String jsonRequest = GSON.toJson(connectedTerminalsRequest);
```
rkewlani: formatting got disturbed here. Please revert this.

src/main/java/com/adyen/model/terminal/TerminalAPIRequest.java
```diff
@@ -73,4 +72,19 @@ public String toString() {
                 "saleToPOIRequest=" + saleToPOIRequest +
                 '}';
     }
+
+    /**
+     * Checks if SaleToAcquirerData is null
+     *
+     * @return true if sale to acquirer data is null
+     */
+    public Boolean isSaleToAcquirerDataModelSet() {
```
martinsrenato: The method returns "true" if there's no SaleToAcquirerData set, so I think the name is misleading.
We could rename it or change it so `SaleToAcquirerData == null` returns false

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -83,4 +82,18 @@ public TerminalAPIResponse sync(TerminalAPIRequest terminalAPIRequest) throws IO
...
+
+    /**
+     * Checks if SaleToAcquirerData is set. If it is not creates an encoded base 64 string with application info.
+     *
+     * @param terminalAPIRequest TerminalAPIRequest
+     * @return TerminalAPIRequest
+     */
+    private static TerminalAPIRequest setSaleToAcquirerDataModel(TerminalAPIRequest terminalAPIRequest) {
+        if (terminalAPIRequest != null && terminalAPIRequest.isSaleToAcquirerDataModelSet()) {
+            SaleToAcquirerDataModel saleToAcquirerDataModel = new SaleToAcquirerDataModel();
```
martinsrenato: This object will be created with ApplicationInfo set by default, but I'm a bit confused on how the other fields of SaleToAcquirerDataModel will be set (metadata, keyValueData, etc). Do we need to cover that as well?

AlexandrosMor: The SaleToAcquirerDataModel with all the other nested fields should be set by the consumer application. If for any reason it is not set then it is created by default with only the application info prefilled.

src/main/java/com/adyen/model/terminal/TerminalAPIRequest.java
msilvagarcia: I'd change the order. Didn't see that in my previous review 😬

AlexandrosMor: I'd changed the order

src/main/java/com/adyen/service/TerminalCloudAPI.java
martinsrenato: Oh I see. And we're assuming it will already be set as a base64 string, right?

AlexandrosMor: Yes as it is mentioned [here](https://docs.adyen.com/point-of-sale/pass-application-information#base64-the-application-information)

martinsrenato: Cool, thanks for the explanation :)

src/main/java/com/adyen/service/TerminalCloudAPI.java
```diff
@@ -83,4 +82,18 @@ public TerminalAPIResponse sync(TerminalAPIRequest terminalAPIRequest) throws IO
...
+     *
+     * @param terminalAPIRequest TerminalAPIRequest
+     * @return TerminalAPIRequest
+     */
+    private static TerminalAPIRequest setSaleToAcquirerDataModel(TerminalAPIRequest terminalAPIRequest) {
+        if (terminalAPIRequest != null && !terminalAPIRequest.isSaleToAcquirerDataModelSet()) {
+            SaleToAcquirerDataModel saleToAcquirerDataModel = new SaleToAcquirerDataModel();
+            terminalAPIRequest.getSaleToPOIRequest().getPaymentRequest().getSaleData().setSaleToAcquirerData(saleToAcquirerDataModel.toBase64());
+        }
+        return terminalAPIRequest;
```
rikterbeek: what if is is manually set by the merchant ? do we then not overwrite it or add it with our library info ?

AlexandrosMor:
This approach is not forcing the application info only adding if it is not there in the sale data as base 64. We close it as it is covering what we expect

PR #264, opened 2019-11-01
[PW-1630]: Deprecated idealIssuer field 

rkewlani:
Deprecated idealIssuer and added new field issuer

### DIFF
--- a/src/main/java/com/adyen/model/checkout/DefaultPaymentMethodDetails.java
+++ b/src/main/java/com/adyen/model/checkout/DefaultPaymentMethodDetails.java
@@ -54,8 +54,14 @@ public class DefaultPaymentMethodDetails implements PaymentMethodDetails {
     private String recurringDetailReference;
     @SerializedName("storeDetails")
     private Boolean storeDetails;
+    /**
+     * @deprecated This field is deprecated and new field added {@link DefaultPaymentMethodDetails#issuer } which is  more generic for other payment methods.
+     */
+    @Deprecated
     @SerializedName("idealIssuer")
     private String idealIssuer;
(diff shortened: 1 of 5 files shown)


src/main/java/com/adyen/model/checkout/DefaultPaymentMethodDetails.java
```diff
@@ -258,6 +270,19 @@ public DefaultPaymentMethodDetails idealIssuer(String idealIssuer) {
         return this;
```
Aleffio: should we deprecate this one as well?

src/main/java/com/adyen/model/checkout/DefaultPaymentMethodDetails.java
```diff
@@ -287,38 +312,41 @@ public boolean equals(Object o) {
...
-                && Objects.equals(personalDetails, that.personalDetails)
+                && Objects.equals(personalDetails,
+                                  that.personalDetails)
                 && Objects.equals(encryptedCardNumber, that.encryptedCardNumber)
                 && Objects.equals(encryptedExpiryMonth, that.encryptedExpiryMonth)
                 && Objects.equals(encryptedExpiryYear, that.encryptedExpiryYear)
                 && Objects.equals(encryptedSecurityCode, that.encryptedSecurityCode)
                 && Objects.equals(recurringDetailReference, that.recurringDetailReference)
                 && Objects.equals(storeDetails, that.storeDetails)
-                && Objects.equals(idealIssuer, that.idealIssuer)
```
Aleffio: if we keep this field on the hashCode method, should we keep it here as well?

PR #277, opened 2019-12-03
[PW-1774]Implement SaleToAcquirerDataModel for sending SaleToAcq…

rikterbeek:
…uirerData in the right format to Adyen.

**Description**

**Tested scenarios**

**Fixed issue**:


src/main/java/com/adyen/serializer/SaleToAcquirerDataModelAdapterSerializer.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.adyen.model.terminal.SaleToAcquirerDataModel;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+import java.lang.reflect.Type;
+
+public class SaleToAcquirerDataModelAdapterSerializer implements JsonSerializer<SaleToAcquirerDataModel> {
```
rkewlani: Any specific reason to mention Adapter in name of class ?

src/main/java/com/adyen/model/nexo/SaleData.java
```diff
@@ -75,7 +79,8 @@
      * The Sale to acquirer data.
      */
     @XmlElement(name = "SaleToAcquirerData")
-    protected String saleToAcquirerData;
+    @JsonAdapter(SaleToAcquirerDataModelAdapterSerializer.class)
+    protected SaleToAcquirerDataModel saleToAcquirerData;
```
rkewlani: We can skip Model in name of class and keep it  SaleToAcquirerData because package name already has "model"

src/main/java/com/adyen/model/nexo/SaleData.java
```diff
@@ -75,7 +79,8 @@
      * The Sale to acquirer data.
      */
     @XmlElement(name = "SaleToAcquirerData")
-    protected String saleToAcquirerData;
```
rkewlani: Removing/changing type saleToAcquirerData field means release won't be backward compatible.

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
```diff
@@ -0,0 +1,210 @@
...
+                ", currency='" + currency + '\'' +
+                ", applicationInfo=" + applicationInfo +
+                ", tenderOption='" + tenderOption + '\'' +
+                ", additionalData=" + additionalData +
+                '}';
+    }
+
+    public String toBase64() {
+        String json = PRETTY_PRINT_GSON.toJson(this);
+        return new String(Base64.encodeBase64(json.getBytes()));
```
rkewlani: json.getBytes may return in null pointer exception.

src/main/java/com/adyen/serializer/SaleToAcquirerDataModelAdapterSerializer.java
rikterbeek: not particular I will remove the adapter part.

src/main/java/com/adyen/model/nexo/SaleData.java
rikterbeek: good point I agreed

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
rikterbeek: I am wondering when this could happen. this is always an object with application info so it should always contain data.

src/main/java/com/adyen/model/nexo/SaleData.java
rikterbeek: correct I don't see any other way.

src/main/java/com/adyen/model/terminal/SaleToAcquirerDataModel.java
rkewlani: "this" won't be null but it is possible that conversion to json has some problem and returns in empty json string.

PR #281, opened 2019-12-23
Adding travis_wait to mvn install

FanJups:
**Description**
Why ?
Building such big APIs as Adyen APIs   can take a lot of time.  **It's possible to get a time out.**
[Here](https://github.com/4hwc/4HWCAutonomousCar/commit/96c6396d444f0364150241c5bed3e0b9fc7ba5c8) , I got a time out due to generating javadoc. 
Then, I decided to add **travis_wait** and it works pretty well.

Official docs : https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ jdk:
   - oraclejdk8
   - openjdk7
 
-install: mvn install -Dgpg.skip=true
+#travis_wait : avoid time out
+install: travis_wait mvn install -Dgpg.skip=true
 
 sudo: false
 


rkewlani:
Thanks @FanJups for the PR and suggestion.
We would like to know more details on this because we don't get timeout on this. We never got this issue and it is not reported by others as well. Can you please explain what all steps you are performing so that we can reproduce it at our end.
Thanks,
Rajni 
Adyen

FanJups:
@rkewlani ,  my PR is based on what I faced  while generating Javadoc weeks ago. I thought it would be helpful to prevent such case.

rikterbeek:
Hi @FanJups,

Thanks for this in general it is indeed a good idea to do this with calls that take time to produce the output. @rkewlani more info here https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received We will discuss in the team if we see any negatives if not we will merge it.

Regards,
Rik
Adyen

rikterbeek:
Hi @FanJups,

Discussed in the team and the build should not take more then 10 minutes and if it does we want to optimize this to make it shorter as it is a lightweight repo. We are building the project always for releases and did not have any issues. Could you investigate on your end what is going wrong with the generating of Javadoc as this should be quicker.

Regards,
Rik
Adyen

PR #286, opened 2020-01-13
Release 4.0.1

rkewlani:
**Description**

**Tested scenarios**

**Fixed issue**:


upgrade_release_version.sh
```diff
@@ -0,0 +1,54 @@
...
+         n)
+          NEWVERSION=${OPTARG};;
+    esac
+done
+
+echo "Old release version is" $OLDVERSION
+echo "New release version is"$NEWVERSION
+
+if [ -z "${OLDVERSION}" ]; then
+    echo "Please specify the 'oldversion' argument";
```
rikterbeek: I would not call it old version but current version (with a space in between)

upgrade_release_version.sh
```diff
@@ -0,0 +1,54 @@
...
+    usage;
+    exit 1;
+fi
+
+if [ -z $NEWVERSION ]; then
+    echo "Please specify the 'newversion' argument";
```
rikterbeek: add space between new and version.

upgrade_release_version.sh
rkewlani: this is argument name so there is no space here.

upgrade_release_version.sh
rkewlani: same here. it is argument name.

upgrade_release_version.sh
```diff
@@ -0,0 +1,54 @@
...
+        o)
+          OLDVERSION=${OPTARG};;
+
+         n)
+          NEWVERSION=${OPTARG};;
+    esac
+done
+
+echo "Old release version is" $OLDVERSION
+echo "New release version is"$NEWVERSION
```
Aleffio:
```suggestion
echo "New release version is" $NEWVERSION
```

Issue #284, opened 2020-01-10
API Idempotency transient errors HTTP header


alansegar-hb:
**Java version**: jdk-11.0.4.11
**Library version**: 2.7.2
**Description**
When implementing retries for payment requests using the Java library as follows:

        RequestOptions requestOptions = new RequestOptions();
        requestOptions.setIdempotencyKey(uuid.toString());
        log.info("Calling Adyen with Idempotency-Key uuid {}", uuid.toString());
        PaymentsResponse paymentsResponse = checkout.payments(paymentsRequest, requestOptions);

https://docs.adyen.com/development-resources/api-idempotency#transient-errors
describes the following:

> HTTP header Transient-Error with the value true, which indicates that the request can be retried at a later time using the same idempotency key. If the API does not return a transient error header, or returns a header with a value of false, do not retry the request.

Is there a way to access this Transient-Error header / boolean value in the Java library? 
I've had a look in `PaymentsResponse`, `ServiceError` and `ApiException` but can't find it.

`HTTPClientException` has a `getResponseHeaders` method which may give access to this information, but I don't have access to that type of exception in the checkout.payments call.

Internally that calls Resource.request which has the following catch:
```
} catch (HTTPClientException e) {
    responseBody = e.getResponseBody();
    apiException = new ApiException(e.getMessage(), e.getCode());
}
```
so the `ApiException` that I get back doesn't allow me to get to the  headers.

AlexandrosMor:
Hello @alansegar-hb,

Thank you for raising this issue. We are going to investigate it and come back to you as soon as possible.

kind regards,
Alexandros
Adyen

PR #288, opened 2020-01-21
[PW-2034] Add http response headers to ApiException

AlexandrosMor:
**Description**
This story is about to access the Transient-Error header value through http client exception. The response returns an HTTP 409 – Conflict status with the error code 704: "request already processed or in progress" but not any http headers.
For information about the Transient-Error check [docs](https://docs.adyen.com/development-resources/api-idempotency#transient-errors)

**Tested scenarios**
Internal e2e test /payments making two transactions with the same idempotency key.

**Fixed issue**:   #284

### DIFF
--- a/src/main/java/com/adyen/service/exception/ApiException.java
+++ b/src/main/java/com/adyen/service/exception/ApiException.java
@@ -20,6 +20,8 @@
  */
 package com.adyen.service.exception;
 
+import java.util.List;
+import java.util.Map;
 import com.adyen.model.ApiError;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/adyen/service/exception/ApiException.java
```diff
@@ -53,6 +63,9 @@ public void setStatusCode(int statusCode) {
         this.statusCode = statusCode;
     }
 
+    public Map<String, List<String>> getResponseHeaders() {
+        return responseHeaders;
+    }
     @Override
     public String toString() {
         return "ApiException{" + "error=" + error + ", statusCode=" + statusCode + ", message=" + getMessage() + '}';
```
martinsrenato: minor: add responseHeaders in the toString method

PR #308, opened 2020-02-02
Update BarcodeOuput to reflect API doc

pc-coholic:
**Description**
This PR addresses two issues:
- Changing the BarcodeType-ENUM from `QRCODE` to `QRCode`
- Changing the barcode's value to `barcodeValue`

**Tested scenarios**
None, since the only terminal available to me, the V400m, apparently doesn't support `DisplayRequest` calls.

However, these fixes take the error messages from "Unexpected field value" and "field barcodeValue missing" to "Unsupported output format" - which seems like an improvement?

**Fixed issue**:  No issue opened


AlexandrosMor:
Hello @pc-coholic,

Thank you for opening this PR, Is it possible to add a unit test related to this change? 

Kind regards,
Alexandros

PR #323, opened 2020-03-25
PW-1705

andrei-dediu:
Added the store field to the PaymentsRequest and PaymentMethodsRequest for the Checkout API.

**Description**

Added the store property field to the PaymentsRequest and PaymentMethodsRequest manually as it was requested for the ticket. The OpenAPI models have not yet been updated with these fields so we can not generate them yet.

**Tested scenarios**

None.

**Fixed issue**: 

PW-1705

### DIFF
--- a/src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
+++ b/src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
@@ -63,7 +63,7 @@ public class PaymentMethodsRequest {
     @SerializedName("accountInfo")
     private AccountInfo accountInfo = null;
 
-    @SerializedName("applciationInfo")
+    @SerializedName("applicationInfo")
     private ApplicationInfo applicationInfo = null;
 
     @SerializedName("configId")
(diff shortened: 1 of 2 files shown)


src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
```diff
@@ -87,6 +87,9 @@
     @SerializedName("threeDSAuthenticationOnly")
     private Boolean threeDSAuthenticationOnly = null;
 
+    @SerializedName("store")
+    private String store = null;
```
rkewlani: Please update equals() and toString() methods to add this field

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
```diff
@@ -185,6 +185,9 @@
     @SerializedName("storePaymentMethod")
     private Boolean storePaymentMethod = null;
 
+    @SerializedName("store")
+    private String store = null;
```
rkewlani: Please update equals()  method to add this field

src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
andrei-dediu: Nice catch, updated.
Also, the applicationInfo serialised annotation was mispelled, updated that as well.

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
andrei-dediu: Also updated !

PR #324, opened 2020-03-30
Improve minor code changes

smic1909:
**Description**
Small code improvements

**Tested scenarios**
No additional tests needed since no logic has changed.

**Fixed issue**
This change is not related to any created Github Issue

### DIFF
--- a/src/main/java/com/adyen/model/checkout/InputDetail.java
+++ b/src/main/java/com/adyen/model/checkout/InputDetail.java
@@ -24,10 +24,7 @@ package com.adyen.model.checkout;
 
 import com.google.gson.annotations.SerializedName;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
 /**
  * InputDetail
@@ -64,6 +61,11 @@ public class InputDetail {
     }
 
     public InputDetail putConfigurationItem(String key, String configurationItem) {
+
+        if (configuration == null) {
+            configuration = new HashMap<>();
+        }
+
         configuration.put(key, configurationItem);
         return this;
     }
--- a/src/main/java/com/adyen/model/checkout/InputDetail.java
+++ b/src/main/java/com/adyen/model/checkout/InputDetail.java
@@ -24,7 +24,11 @@ package com.adyen.model.checkout;
 
 import com.google.gson.annotations.SerializedName;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * InputDetail
--- a/src/main/java/com/adyen/model/checkout/InputDetail.java
+++ b/src/main/java/com/adyen/model/checkout/InputDetail.java
@@ -64,12 +64,7 @@ public class InputDetail {
     }
 
     public InputDetail putConfigurationItem(String key, String configurationItem) {
-
-        if (this.configuration == null) {
-            this.configuration = null;
-        }
-
-        this.configuration.put(key, configurationItem);
+        configuration.put(key, configurationItem);
         return this;
     }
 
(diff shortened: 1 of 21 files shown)


rikterbeek:
Hi @smic1909,

Many thanks for the Pull Request! A lot of code is generated by generator so makes perfect sense to clean this code. As it does have a lot of changes we need to be carefull and as well test all our e2e tests so this will take some time. When I manually review it I do see one major concern would be great if you can fix that.

Regards,
Rik
Adyen

src/main/java/com/adyen/model/checkout/InputDetail.java
```diff
@@ -64,12 +64,7 @@ public InputDetail configuration(Map<String, String> configuration) {
     }
 
     public InputDetail putConfigurationItem(String key, String configurationItem) {
-
-        if (this.configuration == null) {
-            this.configuration = null;
```
martinsrenato, line 69: I think this line was wrong before and should be fixed, not removed.
If configuration is null, a new map should be created so it prevents nullPointerExceptions to happen on the next lines
`configuration = new HashMap<>();`

src/main/java/com/adyen/model/checkout/PaymentMethod.java
```diff
@@ -87,12 +87,7 @@ public PaymentMethod configuration(Map<String, String> configuration) {
     }
 
     public PaymentMethod putConfigurationItem(String key, String configurationItem) {
-
-        if (this.configuration == null) {
-            this.configuration = null;
```
martinsrenato, line 92: I think this line was wrong before and should be fixed, not removed (same case as in InputDetail)

src/main/java/com/adyen/model/checkout/RecurringDetail.java
```diff
@@ -95,12 +95,7 @@ public RecurringDetail configuration(Map<String, String> configuration) {
     }
 
     public RecurringDetail putConfigurationItem(String key, String configurationItem) {
-
-        if (this.configuration == null) {
-            this.configuration = null;
```
martinsrenato, line 100: Same as other putConfigurationItem methods

src/main/java/com/adyen/model/checkout/Redirect.java
```diff
@@ -49,12 +49,7 @@ public Redirect data(Map<String, String> data) {
     }
 
     public Redirect putDataItem(String key, String dataItem) {
-
-        if (this.data == null) {
-            this.data = null;
```
martinsrenato, line 54: Same as the putConfigurationItem case

src/main/java/com/adyen/model/checkoututility/OriginKeysResponse.java
```diff
@@ -41,12 +41,7 @@ public OriginKeysResponse originKeys(Map<String, String> originKeys) {
     }
 
     public OriginKeysResponse putOriginKeysItem(String key, String originKeysItem) {
-
-        if (this.originKeys == null) {
-            this.originKeys = null;
```
martinsrenato, line 46: Same as the putConfigurationItem case

src/main/java/com/adyen/model/checkout/InputDetail.java
smic1909: Fixed.

src/main/java/com/adyen/model/checkout/PaymentMethod.java
smic1909: Fixed.

src/main/java/com/adyen/model/checkout/RecurringDetail.java
smic1909: Fixed.

src/main/java/com/adyen/model/checkout/Redirect.java
smic1909: Fixed.

src/main/java/com/adyen/model/checkoututility/OriginKeysResponse.java
smic1909: Fixed.

smic1909:
> Hi @smic1909,
> 
> Thanks for fixing the issues reported on comments. The methods are functioning correctly now.
> It seems, however, that your changes added some wildcard imports ( **'*'** ) on some classes, and that causes a checkstyle failure on the [Travis CI build](https://travis-ci.org/github/Adyen/adyen-java-api-library/builds/670269637).
> Could you please revert that and add the specific imports needed?
> 
> Regards,
> Renato
> Adyen

Sure. Replaced the star import. Please validate.

PR #332, opened 2020-04-16
Adds phone number to AccountHolderDetails

viZu:
**Description**
The java api library was not compatible anymore with the
web api since the web api does return a phone number object
instead of the fullPhoneNumber field.

**Tested scenarios**
No new test scenarios since it only adds a field and getter/setter to the AccountHolderDetails

**Fixed issue**:


martinsrenato:
Hi @viZu ,

I just merged your changes, they will be included on our next release, which should happen soon.
Thanks again for the contribution!

Regards,
Renato
Adyen

PR #343, opened 2020-05-06
PW-2289: Add paymentMethodDetails classes

Aleffio:
**Description**

**Tested scenarios**

**Fixed issue**:


src/test/java/com/adyen/CheckoutTest.java
```diff
@@ -417,6 +405,176 @@ public void TestSepaPaymentMethodDetails() {
...
+        SepaDirectDebitDetails sepaDirectDebitDetails = new SepaDirectDebitDetails();
+        sepaDirectDebitDetails.setOwnerName("A. Schneider");
+        sepaDirectDebitDetails.setIban("DE87123456781234567890");
+
+        PaymentsRequest paymentsRequest = createPaymentsCheckoutRequest();
+        paymentsRequest.setPaymentMethod(sepaDirectDebitDetails);
+
+        String jsonRequest = PRETTY_PRINT_GSON.toJson(paymentsRequest);
+
+        assertEquals("{\n"
```
andrei-dediu, line 419: Did we always use hardcoded values ? Didn't we have a file with the results to compare against ?

Aleffio: for the requests we are using hardcoded values, for the responses we are using .json files, I can create files for requests as well

andrei-dediu: Alright, if that's how it is overall then okay by me.

rikterbeek:
> I was thinking if we could not change `DefaultPaymentMethodDetails` (maybe even deprecate it) and the new classes all implent `PaymentMethodDetails`. Than there would be no need for the new interface `OneOfPaymentRequestPaymentMethod`. Do you think that would work?

we should use the already existing interface called PaymentMethodDetails and remove the onOff.. interface so we are backwards compatible :)

src/main/java/com/adyen/model/checkout/AchDetails.java
```diff
@@ -0,0 +1,209 @@
...
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+/**
+ * AchDetails
+ */
+
+public class AchDetails implements OneOfPaymentRequestPaymentMethod {
```
rikterbeek, line 32: use the PaymentMethodDetails interface.

src/main/java/com/adyen/model/checkout/PaymentsRequest.java
```diff
@@ -114,7 +114,7 @@
     @SerializedName("orderReference")
     private String orderReference = null;
     @SerializedName("paymentMethod")
-    private PaymentMethodDetails paymentMethod = null;
+    private OneOfPaymentRequestPaymentMethod paymentMethod = null;
```
rikterbeek, line 117: keep the PaymentMethodDetails so it is backwards compatible.

src/main/java/com/adyen/model/checkout/PaymentMethodDetails.java
```diff
@@ -21,6 +21,5 @@
 package com.adyen.model.checkout;
 
 public interface PaymentMethodDetails {
```
andrei-dediu, line 23: I do not see the point of an empty interface just for the sake of convenience. I suggest the following:

- Make it a generic interface to support the different types of the property "type", so we can still keep the required methods to be implemented.
- Add comments that give a hint to the minimum required methods to be created by a class implementing this interface.

Otherwise this interface does not have much of a purpose, from an interface point of view.

src/main/java/com/adyen/model/checkout/PayPalDetails.java
```diff
@@ -84,7 +84,7 @@ public SubtypeEnum read(final JsonReader jsonReader) throws IOException {
     }
 
     @SerializedName("subtype")
-    private SubtypeEnum subtype = null;
+    private SubtypeEnum subtype = SubtypeEnum.REDIRECT;
```
rikterbeek, line 87: why the default redirect would keep this optional to the user.

src/main/java/com/adyen/model/checkout/PaymentMethodDetails.java
rikterbeek: can't we add the type as a default get/setter to implement ?

Aleffio: some classes have enums as type, so we cannot use String in the interface

andrei-dediu: Yeah we can add empty default methods as a hint, but then when implementing the interface you would have to override your getters/setters as to not use the default ones, which is a bit counter-intuitive in my opinion.

rikterbeek:
Could you move the details grouped together in a folder that helps for visibility of the paymentMethods that we have.

src/main/java/com/adyen/model/checkout/details/ApplePayDetails.java
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * ApplePayDetails
+ */
+
+public class ApplePayDetails implements PaymentMethodDetails {
+    /**
+     * Possible types
+     */
+    public static final String APPLE_PAY = "applepay";
```
rkewlani, line 43: constant name APPLEPAY is more appropriate

Aleffio: Apple Pay payment method is two words, while Dotpay is a single one

Aleffio: changed to APPLEPAY for consistency with txvariant

src/main/java/com/adyen/model/checkout/details/DragonpayDetails.java
```diff
@@ -0,0 +1,143 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+/**
+ * DragonpayDetails
+ */
+
+public class DragonpayDetails implements PaymentMethodDetails {
+    @SerializedName("issuer")
```
rikterbeek, line 34: should we not have a constant here ?

src/main/java/com/adyen/model/checkout/details/LianLianPayDetails.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+/**
+ * LianLianPayDetails
+ */
+
+public class LianLianPayDetails implements PaymentMethodDetails {
+    @SerializedName("telephoneNumber")
```
rikterbeek, line 34: Do we not need a constant here ?

src/main/java/com/adyen/model/checkout/details/MbwayDetails.java
```diff
@@ -0,0 +1,143 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+/**
+ * MbwayDetails
+ */
+
+public class MbwayDetails implements PaymentMethodDetails {
+    @SerializedName("shopperEmail")
```
rikterbeek, line 34: do we not need a constant here ?

src/main/java/com/adyen/model/checkout/details/MolPayDetails.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+/**
+ * MolPayDetails
+ */
+
+public class MolPayDetails implements PaymentMethodDetails {
+    @SerializedName("issuer")
```
rikterbeek, line 34: Do we not need a constant here ?

rkewlani: no, because this is name of the field and used only one. It is present in every model class and we will end up in tons of constants if we go for it.

src/main/java/com/adyen/model/checkout/details/DragonpayDetails.java
rkewlani: no, because this is name of the field and used only one. It is present in every model class and we will end up in tons of constants if we go for it.

src/main/java/com/adyen/model/checkout/details/LianLianPayDetails.java
rkewlani: same as above.

src/main/java/com/adyen/model/checkout/details/MbwayDetails.java
rkewlani: same as above.

src/main/java/com/adyen/model/checkout/details/MolPayDetails.java
rkewlani: same as above.

PR #347, opened 2020-05-14
[PW-2297] Fix MenuEntryNumber

martinsrenato:
**Description**
Change MenuEntryMenuEntry to array

**Tested scenarios**
Added test with an example response for a InputRequest

### DIFF
--- a/src/test/java/com/adyen/TerminalCloudAPITest.java
+++ b/src/test/java/com/adyen/TerminalCloudAPITest.java
@@ -81,7 +81,7 @@ public class TerminalCloudAPITest extends BaseTest {
         TerminalCloudAPI terminalCloudApi = new TerminalCloudAPI(client);
 
         TerminalAPIRequest terminalAPIPaymentRequest = createTerminalAPIPaymentRequest();
-        
+
         // add some data
         SaleToPOIRequest saleToPOIRequest = new SaleToPOIRequest();
         PaymentRequest paymentRequest = new PaymentRequest();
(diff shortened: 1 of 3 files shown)


src/test/java/com/adyen/TerminalCloudAPITest.java
```diff
@@ -189,4 +189,23 @@ public void syncAbortRequestSuccess() throws Exception {
...
+     * Test success flow for Input Request on POST /sync
+     */
+    @Test
+    public void syncInputRequestSuccess() throws Exception {
+        Client client = createMockClientFromFile("mocks/terminal-api/input-request-success.json");
+        TerminalCloudAPI terminalCloudApi = new TerminalCloudAPI(client);
+
+        TerminalAPIResponse requestResponse = terminalCloudApi.sync(new TerminalAPIRequest());
+
+        assertNotNull(requestResponse);
```
andrei-dediu, line 203: Is the success for this test determined by the fact that none of the properties are null ?

martinsrenato: The point is to check if we get a MenuEntryNumber.
I usually like to also add null checks before that, but I can remove if you guys feel it's not needed.

Aleffio:
can you please rebase on develop? it seems the history is a bit off

src/test/java/com/adyen/TerminalCloudAPITest.java
andrei-dediu: Nah was just checking, it's alright for me. It can be useful if we ever move to a different source instead of the mocked ones.

PR #350, opened 2020-05-18
Add java 11 to travis-ci

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -187,5 +187,20 @@
             <version>2.28.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.3.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.sun.xml.bind</groupId>
+            <artifactId>jaxb-core</artifactId>
+            <version>2.3.0</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


martinsrenato:
https://travis-ci.org/github/Adyen/adyen-java-api-library/jobs/688345740

This actually fails on travis, because the nexo models use some dependencies that were removed from Java 11. (check https://stackoverflow.com/questions/52502189/java-11-package-javax-xml-bind-does-not-exist)

It should work if we add the dependencies back via maven
```
<dependency>
  <groupId>javax.xml.bind</groupId>
  <artifactId>jaxb-api</artifactId>
  <version>2.3.0</version>
</dependency>
<dependency>
  <groupId>com.sun.xml.bind</groupId>
  <artifactId>jaxb-core</artifactId>
  <version>2.3.0</version>
</dependency>
<dependency>
  <groupId>com.sun.xml.bind</groupId>
  <artifactId>jaxb-impl</artifactId>
  <version>2.3.0</version>
</dependency>
```
(maybe not all of them are needed)

pom.xml
```diff
@@ -187,5 +187,20 @@
             <version>2.28.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
rkewlani, line 190: We are just changing jdk version. Do we need additional dependencies in java11 ?

PR #355, opened 2020-05-22
Feature/code cleanup

smic1909:
**Description**
Make use of Java 8 language
- Remove redundant local variable
- Remove unnecessary conversion
- rename packagename `Util` to `util`
- Replace `null` check with `Objects.isNull()`

**Fixed issue**:  
Follow up on #333

### DIFF
--- a/src/main/java/com/adyen/Config.java
+++ b/src/main/java/com/adyen/Config.java
@@ -22,6 +22,8 @@ package com.adyen;
 
 import com.adyen.enums.Environment;
 
+import java.util.Objects;
+
 public class Config {
     protected String username;
     protected String password;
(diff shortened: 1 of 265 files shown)


src/main/java/com/adyen/Config.java
```diff
@@ -143,7 +145,7 @@ public void setHmacKey(String hmacKey) {
     }
 
     public String getCheckoutEndpoint() {
-        if (checkoutEndpoint == null || checkoutEndpoint.isEmpty()) {
+        if (Objects.isNull(checkoutEndpoint) || checkoutEndpoint.isEmpty()) {
```
andrei-dediu, line 148: Hey @smic1909 

I find the Objects.isNull a bit redundant when it's not being used as a predicate in a stream context. It's just object == null on the inside as well. If you agree, could you restore these one back to the regular null check.

smic1909:
All changes will be listed in #356.

PR #356, opened 2020-05-22
Feature/new constants

smic1909:
**Description**
Make use of Util.toIndentedString method 
- replace duplicate code 
- replace `\n` with constant

Ideally this PR will be merged after #355 since this branch has been created on the named one


smic1909:
The idea was to separate these big changes into two PRs, that is why I added the comment to the description to first merge #355. I will revert the `Objects.isNull` changes as you suggested and close #355.

checkstyle.xml
```diff
@@ -146,7 +146,7 @@
         <module name="EqualsHashCode"/>
         <!--<module name="HiddenField"/>-->
         <module name="IllegalInstantiation"/>
-        <module name="InnerAssignment"/>
+        <!-- <module name="InnerAssignment"/>-->
```
andrei-dediu, line 149: Hey @smic1909 

What's the reasoning between commenting this check ?

smic1909: Good catch. This was needed for some changes I did locally. Reverted.

smic1909:
@andrei-dediu Will this PR be merged? It is already open for a while and I wonder why your the pending reviewer did not check the PR yet.

PR #367, opened 2020-06-25
revert null, line_break and line_break_with_whitespace constants

Aleffio:
**Description**

**Tested scenarios**

**Fixed issue**:


smic1909:
@Aleffio May I ask for the reason of the revert? I would like to support here in case something was not working as expected.

andrei-dediu:
Heu @smic1909,

We've reverted these changes after discussing internally that they were not all that beneficial at the moment. It also ands a little bit of extra work for us every time we import new models as the generator generates the models with that structure.

We appreciate your efforts, next time we will try to communicate our doubts properly on the pull review.

Regards,
Andrei.

PR #417, opened 2020-09-14
Add remark for java 11

achawki:

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Library supports all APIs under the following services:
 
 ## Requirements
 
-* Java 8 or higher
+* Java 8 or higher. Java 11 is not yet supported.
 
 ## Installation
 


nickcaballero:
Question about this - what exactly is not supported under Java 11? And does that mean that versions > Java 11 are also not supported?

achawki:
I cloned the repo, executed the unit tests and got following error:
```
java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlElement
```
Hence I mentioned java 11 is not supported. Not sure about other versions.

Closing PR since it might create more confusion.

nickcaballero:
@achawki I see. That makes sense - JAXB is no longer part of Java 11. Maybe it would make more sense to add the JAXB libraries as a dependency here? https://stackoverflow.com/a/52502208

achawki:
That would definitely make sense. There may, however be more adaptations necessary.

PR #422, opened 2020-09-16
PW-3018 Added order and splitCardFundingSources to PaymentMethodsRequest

VladimirButch:
**Description**
Added order and splitCardFundingSources to PaymentMethodsRequest

**Tested scenarios**
Just used existing test scenario since we are not using any fields except merchantAccount

### DIFF
--- a/src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
+++ b/src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
@@ -298,6 +298,7 @@ public class PaymentMethodsRequest {
 
     /**
      * Choose if a specific transaction should use the Real-time Account Updater, regardless of other settings.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/com/adyen/model/checkout/PaymentMethodsRequest.java
```diff
@@ -319,6 +325,34 @@ public void setThreeDSAuthenticationOnly(Boolean threeDSAuthenticationOnly) {
         this.threeDSAuthenticationOnly = threeDSAuthenticationOnly;
     }
 
+    public PaymentMethodsRequest order(CheckoutOrder order) {
+        this.order = order;
+        return this;
+    }
+
```
rkewlani, line 332: minor: multiple blank lines in this file.

rkewlani (dismissed):

Issue #340, opened 2020-04-28
[PW-2415] Enums in models POJOs does not work with Jackson


deepu105:
**Describe the bug**
When using Jackson to serialize the POJOs the enums get the toString rather than value, this makes the POJOs unusable directly and requires a bridging DTO that just remaps with Jackson annotations

**How to reproduce**
Try to use the POJO with a webapp that uses Jackson as the serializer, for example, any spring applications

**Expected behavior**
POJOs should be technology agnostic

**Desktop (please complete the following information):**
 - Java Version: 11
 - Library Version: 5.0.0

deepu105:
I would recommend also adding the Jackson annotations to the POJO as Jackson is more popular the Gson anyway. The annotations can coexist without any issues

see https://stackoverflow.com/questions/32958521/can-gson-and-jackson-annotations-coexist-in-the-same-pojo

andrei-dediu:
Hey @deepu105 

Thanks for opening the issue and the additional suggestion. I have looked at our Java API Library code and indeed and there are some enums with different values from their name. I created a ticket to adjust the serializing and deserializing process by either seeing if we can further customise the process with Gson or by implementing the Jackson annotations.

Regards,
Andrei.

saquibsayyad:
Jackson annotations has been added in the POJO's, will be available in the next release.

PR #423, opened 2020-09-17
PW-2415 Added Jackson annotations in models for serialization

saquibsayyad:
**Description**
Added Jackson annotations in the models so that they can be serialized the same way as GSON

**Fixed issue**:  
#340


andrei-dediu (dismissed):

pom.xml
```diff
@@ -25,6 +25,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <jackson.version>2.11.2</jackson.version>
```
rkewlani, line 28: can you please use 2.9.7?

saquibsayyad: done

andrei-dediu (approved):

Aleffio (dismissed):

rkewlani (dismissed):

Aleffio (approved):

andrei-dediu (approved):

deepu105:
@saquibsayyad @andrei-dediu seems like annotations are missing for `SepaDirectDebitDetails`, any reason this class is needed as the `DefaultPaymentMethodDetails` seems to have this field. If not required we should mark it as deprecated IMO

saquibsayyad:
Hi @deepu105 ,

`SepaDirectDebitDetails` has field name same as in the json representation, so Jackson annotations won't be required. 
`DefaultPaymentMethodDetails` is part of old implementation, we will be deprecating it in future.

Thanks,
Saquib

deepu105:
But the data sent from the client side still uses the old format so this doesn't work at the moment and I would assume changing how the front end sends data is a breaking change?

saquibsayyad:
Hi @deepu105,

Both the formats are working for now. `sepa.ibanNumber` and `sepa.ownerName` fields are already deprecated internally and will be removed later. It is replaced by `iban` and `ownerName`.
Thanks,
Saquib

PR #426, opened 2020-09-18
Fix filename collision

maxsumrall:
**Description**
Deletes `BilldeskOnlineDetails.java` and `BilldeskWalletDetails.java` which have filenames which cause a collision for case-insensitive filesystems due to identical files with a filename which only differs by casing. 

It looks like this was attempted to be fixed by marking the files with the incorrect name with the `@Deprecated` annotation, but deprecation should not be an issue here and the file can be directly deleted/renamed. 

**Tested scenarios**
N/A

**Fixed issue**: 
It's not possible to have multiple files with the same filename on filesystems which are case-insensitive. In this case, you will have issues checking out the code and creating a correct diff because there is an irreconcilable change to the git project.


maxsumrall:
I was thinking about this over the weekend and had some thoughts.

• If you think this might be a breaking change that causes problems for some users, that would be a reason to not rename the filename _until_ you plan to do a major version release. 
• If so, then I it's still *not* a solution to have two versions of the file in the repo. This is never going to work. 
• You should then leave the file name as-is, and rename the file at the moment you are doing a major version update. 

I'd vote for just renaming the file/delete the old files (e.g. merge this PR). Otherwise, you should delete the "new" files with the correct casing name, and live with the incorrectly cased files, and rename them later when doing a major release. 

I think what's clear is that you have to do something, since now the project is essentially broken.

andrei-dediu (approved):

andrei-dediu:
Hey @maxsumrall,

Thanks for the PR. Yeah, this is going to be a breaking change but indeed it's a blocking issues for case insensitive filesystems. We will go ahead with removing the files and put the appropriate warnings in the coming major release.

Thanks,
Andrei.

Aleffio (approved):

Issue #402, opened 2020-08-26
Crash when using a local certificate


DantheCodingGui:
**Describe the bug**
When making a (Resource) request with a local certificate (`terminalCertificatePath` set on the Client config), I'm getting an `ApiException`. 

Debugged it and found that a call to `FileInputStream` is the issue, `reset` is called on it which throws an `IOException` (mark/reset not supported), (HttpUrlConnectionClient:272 on lib version 8.1.0). Gets rethrown as an `ApiException`.

**Additional context**
Using this through Android, supporting Java 8.

Aleffio:
Hi @DantheCodingGui 
thank you for opening the issue, I am opening a ticket to investigate on that.
@jnorkus did you perhaps encounter the same issue when testing a request with a local certificate on Android?

Regards,
Alessio
Adyen

jnorkus:
Hi. I did not test the `terminalCertificatePath` on Android. I load the stream from a resource which supports resetting and I found it to be needed for that scenario.

saquibsayyad:
We have removed `terminalCertificateStream.reset()` as it fails for the InputStream which does not support mark() (such as FileInputStream) or if no mark is set before calling reset.  Check [reset()](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--) for more details.

PR #430, opened 2020-09-23
PW-2994 InputStream can only be reset if mark is supported

saquibsayyad:
**Description**
InputStream can only be reset if the given stream supports mark and it is set to correct position, otherwise it will throw IOException. Hence we have to assume that the stream is at the start of certificate data.

**Tested scenarios**

**Fixed issue**:  #402

### DIFF
--- a/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
+++ b/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
@@ -269,7 +269,7 @@ public class HttpURLConnectionClient implements ClientInterface {
             try {
                 // Create new KeyStore for the terminal certificate
                 CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
-                terminalCertificateStream.reset();
+
                 X509Certificate cert = (X509Certificate) certificateFactory.generateCertificate(terminalCertificateStream);
 
                 KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());


src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
```diff
@@ -269,7 +269,7 @@ private void installCertificateVerifier(URLConnection connection, InputStream te
             try {
                 // Create new KeyStore for the terminal certificate
                 CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
-                terminalCertificateStream.reset();
```
achawki, line 272: Was there actually a scenario for the `reset`, I couldn't find a `mark`?
If yes, we could use something like `if terminalCertificateStream.markSupported()`.

saquibsayyad: I think it was just done considering that we will always have stream from beginning, but not all streams support reset. Also even if mark is supported but if it is not set, it will throw exception on reset.

achawki (approved):

rkewlani (approved):

jnorkus:
I believe that removing `reset()` introduces a bug when using `Config.terminalCertificateStream` as a reusable certificate stream. The use case is on Android - I load the certificate from resources. I will not be able to update to latest version after these changes.

saquibsayyad:
Hi @jnorkus ,

Thanks for letting us know. I'm working on another fix to change the way we load certificate, so that we don't have to load it every-time but just once.

Thanks,
Saquib

saquibsayyad:
https://github.com/Adyen/adyen-java-api-library/pull/434 is open.

PR #434, opened 2020-09-25
PW-2994 Load terminal certificate in Config

saquibsayyad:
**Description**
Load terminal certificate in Config instead of HttpURLConnectionClient. This is to avoid loading certificate on each http call and instead load it once for all.

Following methods has been changed to `setTerminalCertificate()`:
```
setTerminalCertificatePath(String terminalCertificatePath)
setTerminalCertificateStream(InputStream terminalCertificateStream)
```


Aleffio (approved):

rkewlani (approved):

vuphamdirox:
@saquibsayyad 
Following the code here: https://docs.adyen.com/point-of-sale/design-your-integration/choose-your-architecture/local/protect-with-library?tab=java_3

config.setTerminalCertificate(**"Local path to the Adyen root certificate")**;

I tried to install the the root certificate to the terminal S1F2 successfully.

How can I get the local path of certificate on terminal?

Regards,
Vu

PR #452, opened 2020-10-27
[PW-3275] Add productUrl and imageUrl to LineItem class

zurmalik:
**Description**
Add productUrl and imageUrl to LineItem class

**Tested scenarios**
The payment from postman is accepted in test environment.

### DIFF
--- a/src/test/java/com/adyen/BaseTest.java
+++ b/src/test/java/com/adyen/BaseTest.java
@@ -85,6 +85,8 @@ import static org.mockito.Mockito.when;
 
 public class BaseTest {
     protected static final Gson PRETTY_PRINT_GSON = new GsonBuilder().setPrettyPrinting().create();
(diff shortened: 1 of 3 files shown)


src/test/java/com/adyen/BaseTest.java
```diff
@@ -229,6 +233,8 @@ protected PaymentsRequest createAfterPayPaymentRequest() {
                 .taxAmount(52L)
                 .amountIncludingTax(300L)
                 .taxCategory(LineItem.TaxCategoryEnum.HIGH)
+                .imageUrl("dummy_protocol/image_url/")
```
andrei-dediu, line 236: Why are we using the string here when we have the constant ?

zurmalik: fixed. pushed.

Aleffio (approved):

rkewlani (approved):

PR #455, opened 2020-11-04
PW-2889 Updated CreatePaymentLinkRequest model with checkout specs v66

saquibsayyad:
**Description**
Updated CreatePaymentLinkRequest model with checkout specs v66

**Tested scenarios**
Create payment link with new fields

### DIFF
--- a/src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
+++ b/src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
@@ -14,7 +14,7 @@
  *
  * Adyen Java API Library
  *
- * Copyright (c) 2019 Adyen B.V.
+ * Copyright (c) 2020 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  */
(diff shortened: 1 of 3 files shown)


src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
```diff
@@ -59,24 +79,99 @@
...
     @SerializedName("reference")
     private String reference = null;
 
     @SerializedName("returnUrl")
     private String returnUrl = null;
 
+    @SerializedName("reusable")
+    private Boolean reusable = null;
+
+    @SerializedName("riskData")
```
Aleffio, line 154: the `riskData` field seems to appear only from v65 on the specs

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
```diff
@@ -59,24 +79,99 @@
...
     private String shopperLocale = null;
 
+    @SerializedName("shopperName")
+    private Name shopperName = null;
+
     @SerializedName("shopperReference")
     private String shopperReference = null;
 
+    @SerializedName("splits")
+    private List<Split> splits = null;
```
Aleffio, line 170: can you please update `Split.TypeEnum`? The possible value "BalanceAccount" has been added to the specs.

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
```diff
@@ -50,6 +66,10 @@
     @SerializedName("countryCode")
     private String countryCode = null;
 
+    @SerializedName("deliverAt")
+    @JsonAdapter(DateTimeGMTSerializer.class)
+    private Date deliverAt = null;
```
Aleffio, line 71: we have `expiresAt` as a String, probably `deliverAt` is autogenerated to Date because of the "format" provided on the specs side. Since "type" is still String on the specs, can you double check with e2e if this field works as expected?

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
saquibsayyad: sorry wrong heading, the models are updated from v66

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
saquibsayyad: updated all dependent models too.

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
saquibsayyad: I internally in the core also it is as Date, i have checked with e2e and it works fine.

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
Aleffio: Client.java points to v64, https://github.com/Adyen/adyen-java-api-library/blob/5e2ac0a0ae3fe86077785711e6f9721b660e5404/src/main/java/com/adyen/Client.java#L51
so I am not sure if this field will be considered, can you create a ticket to update checkout to v65? (v66 doesn't appear on our public API explorer yet)

saquibsayyad: Yes, they are new fields so will be skipped in current version. I have created ticket to update checkout to v65.

Aleffio (approved):

andrei-dediu (approved):

PR #467, opened 2020-11-26
Release 11.0.0

### LINKED ISSUE #461: add shopperLocale parameter in GetOnboardingUrlRequest
**Library version**: 10.0.0

**Description**
Can you add "shopperLocale" parameter in GetOnboardingUrlRequest?

Aleffio:
**Description**
**Important note**
This release will break some old implementation because some fields have been removed in the new **Checkout API** version (v65). Therefore we are releasing this update as new major version.
For more info on the specific fields, please check #465 

#462 Add shopperLocale to GetOnboardingUrlRequest, fixes #461 
#463 Custom deserializer for PaymentMethodDetails 
#464 Added response constants for ModificationResult response field.
#465 Update fields for Checkout v65


saquibsayyad (dismissed):

martinsrenato (dismissed):

saquibsayyad (approved):

src/main/java/com/adyen/model/checkout/details/AndroidPayDetails.java
```diff
@@ -33,7 +33,7 @@
  */
 
 public class AndroidPayDetails implements PaymentMethodDetails {
-    private static final String ANDROIDPAY = "androidpay";
+    public static final String ANDROIDPAY = "androidpay";
```
andrei-dediu, line 36: Not blocking, but is this automatically generated ? It seems weird to have it public and then also have the private variable type reference, but I seem to recall that's just the way we generate it. Just want to double check.

andrei-dediu (approved):

src/main/java/com/adyen/model/checkout/details/AndroidPayDetails.java
Aleffio: it's not automatically generated, we had a discussion on how to include the txvariant in the class, it was hardcoded first in the "type", and then we went for the constant approach.
In this release we are changing them from private to public, as we are comparing the type for the deserializers.

PR #503, opened 2021-03-11
Changed matcher.group() parameter to prevent UnsupportedOperationException on some older Android devices.

tjshakespeare:
**Description**

Using the matcher.group() function with the parameter of int instead of String. Stops an 'UnsupportedOperationException' being thrown when used on some Android devices.

 - group(1) returns the common name
 - group(2) returns the value

It seems like the internals of the matcher.group(String) function works out the index before executing the same code as the matcher.group(index). 

Also, checked to make sure that matcher.group(int) was not deprecated before using it, still a supported and used method.

**Tested scenarios**

- Tested changes on Android devices with the OS ranging from 7-10
- Executed the existing test scripts in the TerminalCommonNameValidatorTest class 

**Fixed issue**:

### DIFF
--- a/src/main/java/com/adyen/terminal/security/TerminalCommonNameValidator.java
+++ b/src/main/java/com/adyen/terminal/security/TerminalCommonNameValidator.java
@@ -44,9 +44,9 @@ public final class TerminalCommonNameValidator {
         Matcher matcher = pattern.matcher(name);
         boolean valid = false;
         while (matcher.find() && !valid) {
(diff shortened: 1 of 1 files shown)


Aleffio:
Hi @tjshakespeare,
Thank you for the pull request! We're going to review it and merge it.

Regards,
Alessio
Adyen

Aleffio (approved):

abhilash-adyen (approved):

PR #518, opened 2021-04-11
PW-4309: Add documentation annotations to nexo classes

Aleffio:


src/main/java/com/adyen/model/nexo/CapturedSignature.java
```diff
@@ -66,21 +70,21 @@ public void setAreaSize(AreaSize value) {
...
      * For example, to add a new item, do as follows:
      * <pre>
      *    getSignaturePoint().add(newItem);
      * </pre>
-     *
-     *
-     *
+     * <p>
+     * <p>
+     * <p>
```
saquibsayyad, line 87: I think it is by mistake.

saquibsayyad (approved):

rkewlani (approved):

src/main/java/com/adyen/model/nexo/CapturedSignature.java
Aleffio: this was autogenerated from the jaxb tool, will update this in another ticket 👍

PR #532, opened 2021-04-29
Improving the library compatibility.

OscarTF:
Recovering compatibility with Android 6 M API level 23 and earlier versions.

java.util.Optional was added in API level 24 (Android 7 N).
https://developer.android.com/reference/java/util/Optional


saquibsayyad (dismissed):

saquibsayyad (approved):

romainbenmansour:
Up. 

We started implementing the adyen java api library and we are currently stuck because of the use of Optional. The proposed change does the trick and is pretty insignificant code wise.

Aleffio (approved):

AlexandrosMor (approved):

PR #567, opened 2021-06-08
Use KeyStore.getDefaultType()

ibartj:
Android throws `java.security.KeyStoreException: JKS not found` as it (the Dalvik virtual machine) does not provide JKS key store type. Please, see KeyStore description at https://developer.android.com/reference/java/security/KeyStore

The call to KeyStore.getDefaultType() was used in HttpURLConnectionClient.java until the refactor to AdyenHttpClient.java in 15.0.0 where it was replaced with the JAVA_KEYSTORE constant. Please, keep using KeyStore.getDefaultType().

**See:**
https://github.com/Adyen/adyen-java-api-library/blob/14.0.0/src/main/java/com/adyen/httpclient/HttpURLConnectionClient.java
https://github.com/Adyen/adyen-java-api-library/blob/15.0.0/src/main/java/com/adyen/httpclient/AdyenHttpClient.java

PR #569, opened 2021-06-10
PW-4495 Added support for client certificate authentication

saquibsayyad:
**Description**
Added support for client certificate authentication. (Right now only supported in live)

**Tested scenarios**
Tested in LIVE


src/main/java/com/adyen/httpclient/AdyenHttpClient.java
```diff
@@ -181,35 +183,63 @@ private URI createUri(String endpoint, Map<String, String> params) throws HTTPCl
...
-
-                // Install the terminal certificate trust manager
-                SSLContext sc = SSLContext.getInstance(SSL);
-                sc.init(null, trustManagers, new java.security.SecureRandom());
-
-                SSLConnectionSocketFactory sslConSocFactory = new SSLConnectionSocketFactory(sc, createHostnameVerifier(config.getEnvironment()));
-                httpClientBuilder.setSSLSocketFactory(sslConSocFactory);
-            } catch (GeneralSecurityException | IOException e) {
-                throw new HTTPClientException("Error loading certificate from path", e);
+            KeyStore keyStore = KeyStore.getInstance(JAVA_KEYSTORE);
```
Aleffio, line 203: should we use `KeyStore.getDefaultType` like suggested in #567?
```suggestion
            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
```

saquibsayyad: > should we use `KeyStore.getDefaultType` like suggested in #567?

corrected it, thanks

src/main/java/com/adyen/httpclient/AdyenHttpClient.java
```diff
@@ -69,7 +69,9 @@
 
     private static final String CHARSET = "UTF-8";
     private static final String TERMINAL_CERTIFICATE_ALIAS = "TerminalCertificate";
+    private static final String JAVA_KEYSTORE = "JKS";
```
Aleffio, line 72: I forgot this one :)
```suggestion
```

Aleffio (approved):

src/main/java/com/adyen/httpclient/AdyenHttpClient.java
saquibsayyad: > I forgot this one :)

my bad, i just applied your suggestion using Github UI instead of doing it in code.

Aleffio (dismissed):

src/main/java/com/adyen/Client.java
```diff
@@ -77,6 +81,33 @@ public Client(String username, String password, Environment environment, String
...
+     * Note: Client certificate authentication is only applicable for PAL and Checkout services in LIVE,
+     * Other services will just use API key for authentication.
+     * @param trustStore Trust store containing server certificate
+     * @param clientKeyStore Client Key store containing client certificate and key
+     * @param clientKeyStorePassword Password for client key store
+     * @param apiKey Adyen API Key
+     * @param region Data center region (EU/US/AU), default EU if not provided
+     */
+    public Client(KeyStore trustStore, KeyStore clientKeyStore, String clientKeyStorePassword, String apiKey, Region region) {
+        this();
```
rkewlani, line 95: minor: this(String apiKey, Environment environment) can be used here.

src/test/java/com/adyen/util/CertificateUtilTest.java
```diff
@@ -50,4 +55,19 @@ public void testLoadCertificateFromInputStream() throws CertificateException {
 
         Assert.assertNotNull(certificate);
     }
+
```
rkewlani, line 58: minor: more tests can be added for other negative scenarios like if resource.getPath() is null or keystore type is null etc.

rkewlani (approved):

src/main/java/com/adyen/Client.java
saquibsayyad: thanks, will update it,

src/test/java/com/adyen/util/CertificateUtilTest.java
saquibsayyad: Added null check and more tests.

src/main/java/com/adyen/Client.java
Aleffio: `this(String apiKey, Environment environment)` sets also the other endpoints(marketpay, hpp,...) that do not support client certificate auth.
Do we really want the merchants to use the same client with the certificate, also for "non certificate" endpoints?

saquibsayyad: `this(apiKey, Environment.LIVE)` was just to replace individual `this(),         this.setEnvironment(Environment.LIVE, null); and this.config.setApiKey(apiKey);` so the behaviour is same.

However if we don't want to setup other endpoints then i can remove the `this.setEnvironment(Environment.LIVE, null)` and just set PAL and Checkout endpoints.

Aleffio (approved):

rkewlani (approved):

PR #571, opened 2021-06-14
PW-4930 Add requiredShopperFields field to CreatePaymentLinkRequest

saquibsayyad:
**Description**
Added requiredShopperFields field in  CreatePaymentLinkRequest

**Tested scenarios**
Create payment link flow.


src/main/java/com/adyen/model/checkout/InstallmentOption.java
```diff
@@ -0,0 +1,213 @@
+/*
```
Aleffio, line 1: very minor: could you please add the correct copyright and auto-indent the file from intellij?

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
```diff
@@ -133,7 +130,7 @@ public void write(final JsonWriter jsonWriter, final RecurringProcessingModelEnu
 
             @Override
             public RecurringProcessingModelEnum read(final JsonReader jsonReader) throws IOException {
-                String value = jsonReader.nextString();
+                Object value = jsonReader.nextString();
```
rkewlani, line 133: possible null pointer. why it is changed to Object from String?

src/main/java/com/adyen/model/checkout/InstallmentOption.java
saquibsayyad: will do it, thanks.

src/main/java/com/adyen/model/checkout/CreatePaymentLinkRequest.java
saquibsayyad: seems it was changed in the model generator, i will revert it back.
Regarding NPE for `JsonReader`, I don't think we need to handle it as it is called by GSON internally.

Aleffio (approved):

rkewlani (approved):

PR #577, opened 2021-06-28
Added initial implementation for company related endpoints for Management API

abhilash-adyen:
Added `/companies` Management API endpoints.


src/main/java/com/adyen/model/management/GenerateHmacKeyResponse.java
```diff
@@ -0,0 +1,79 @@
...
+        return Objects.hash(hmacKey);
+    }
+
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("class GenerateHmacKeyResponse {\n");
+
+        sb.append("    hmacKey: ").append(toIndentedString(hmacKey)).append("\n");
```
achawki, line 75: We could use here `MaskUtil` to mask the key

src/main/java/com/adyen/util/Util.java
```diff
@@ -159,4 +160,10 @@ public static String calculateSessionValidity() {
 
         return fmt.format(sessionDate);
     }
+
+    public static String getPaginationFragment() {
```
achawki, line 164: Is the format actually something generic which other APIs are using as well, or specific to the `Management API`.

src/test/resources/mocks/management/companies/list-companies-response.json
```diff
@@ -0,0 +1,29 @@
+{
+  "_links": {
+    "first": {
+      "href": "http://localhost:8080/configurationapi/v1/companies?pageNumber=1&pageSize=10"
```
achawki, line 4: Minor, we could use here the real `Management API` endpoint of the `TEST` environment for the test samples.

achawki (approved):
I assume

src/main/java/com/adyen/model/management/Company.java
```diff
@@ -0,0 +1,176 @@
...
+import com.google.gson.annotations.SerializedName;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import static com.adyen.util.Util.toIndentedString;
+
+public class Company {
+    @SerializedName("_links")
+    @JsonProperty("_links")
+    private Links links = null;
```
saquibsayyad, line 33: Links model has been updated for most of the resources. Can you regenerate it.

saquibsayyad (approved):

src/main/java/com/adyen/util/Util.java
abhilash-adyen: I haven't noticed any usage of pagination in other parts of the library. Perhaps it can be moved to something more centralized later on.

src/main/java/com/adyen/model/management/Company.java
abhilash-adyen: Fixed :)

PR #593, opened 2021-07-21
Release 16.0.0

Aleffio:
**Description**
***Important Note***
This is a major release, because we are upgrading the Apache HTTP Client version to version 5.1, this might create conflicts if you are using an older version of the client in your project.

#579 Updated Apache HTTP Client to version 5.1 to avoid package conflict in Android
#580 add shopperNotificationReference field to DefaultPaymentMethodsDetails
#583 add /getTaxForm endpoint
#518 add annotations to nexo models for terminal api


saquibsayyad (approved):

src/main/java/com/adyen/httpclient/AdyenHttpClient.java
```diff
@@ -104,7 +107,7 @@ public String request(String endpoint, String requestBody, Config config, boolea
     @Override
     public String request(String endpoint, String requestBody, Config config, boolean isApiKeyRequired, RequestOptions requestOptions, ApiConstants.HttpMethod httpMethod, Map<String, String> params) throws IOException, HTTPClientException {
```
achawki, line 108: Not related to the PR, but I think at some point we should consider a Builder

src/main/java/com/adyen/model/nexo/ICCResetData.java
```diff
@@ -31,12 +33,14 @@
     /**
      * The Atr value.
      */
-    @XmlAttribute(name = "ATRValue")
+    @XmlElement(name = "ATRValue")
+    @Schema(description = "Value of the Answer To Reset of a chip card --Rule: if available", minLength = 1, maxLength = 100)
```
achawki, line 37: Shouldn't this be `minItems` since it's an array?
Or `string` with type `format:byte`?

achawki (approved):

src/main/java/com/adyen/model/nexo/ICCResetData.java
Aleffio: good one, will check with devex if they support `minItems` and create an additional ticket 👍

Issue #529, opened 2021-04-26
AccountHolderStatus.events missing / AccountHolderStatus.accountEvents has wrong name


LukasDrP:
**Describe the bug**

AccountHolderStatus (eg used in the ACCOUNT_HOLDER_STATUS_CHANGE-notification as content->newStatus/oldStatus as described here https://docs.adyen.com/api-explorer/#/NotificationService/v3/post/ACCOUNT_HOLDER_STATUS_CHANGE) should have a field "events" but has instead a field "accountEvents". As a result, "accountEvents" is always empty. 

**How to reproduce**

receiven an ACCOUNT_HOLDER_STATUS_CHANGE-Notification (e.g. by onboarding a new Account and setting its tier to 3) and compare the parsed Object with the plain JSON Event. 

**Expected behavior**

The deserialized (Java-) Object should represent the same data as the serialized (JSON) one. 

**Screenshots**

**Desktop (please complete the following information):**
 - Java Version: 11
 - Library Version: 13.1.0

**Additional context**

Aleffio:
Hi @LukasDrP 
thanks for raising the issue, this was addressed in #522
and the fix will be available in the upcoming 14.0.0 release.

Regards,
Alessio
Adyen

LukasDrP:
Hi @Aleffio, 
thanks your quick response. However, the ArrayList gets filled now, but the AccountEvent-Objects in there are always filled with null-values. The notification before getting parsed to an java-object:
```
{
	"eventDate": "2021-04-30T12:50:27+02:00",
	"eventType": "ACCOUNT_HOLDER_STATUS_CHANGE",
	"executingUserKey": "Status Update",
	"live": "false",
	"pspReference": "8816197798196087",
	"content": {
		"accountHolderCode": "671603ec-90b5-4a10-ad96-b95a8eeba8af",
		"oldStatus": {
			"status": "Active",
			"processingState": {},
			"payoutState": {
				"allowPayout": "false"
			},
			"events": []
		},
		"newStatus": {
			"status": "Active",
			"processingState": {
				"disabled": "false",
				"processedFrom": {
					"currency": "EUR",
					"value": 500000
				},
				"tierNumber": 3
			},
			"payoutState": {
				"allowPayout": "false",
				"disabled": "false"
			},
			"events": [
				{
					"AccountEvent": {
						"event": "InactivateAccount",
						"executionDate": "2021-05-30T12:50:27+02:00",
[...]

PR #595, opened 2021-08-04
Revert "Add AccountEventContainer layer for AccountHolderStatus" (fixes #589)

### LINKED ISSUE #589: calling getAccountHolder results in a null object for AccountEvents
**Describe the bug**

After updating the Adyen Java Lib to 15.1 requesting the user status via /getAccountHolder results in a null object for AccountEvents, even if there are any in the response json. It seems like the lib expects an AccountEventContainer which the response can't provide. This bug is possible related to #529. Here the AccountEventContainer is introduced for ACCOUNT_HOLDER_STATUS_CHANGE-notifications

**How to reproduce**
Request getAccountHolder 

**Expected behavior**
Filled AccountEvent-Objects when there are present in the response json.

** System **
* Java 11
* Adyen Lib 15.1

**Additional context**
Perhabs related to the fix introduced with #529

RickWieman:
This reverts commit 78129f976fa26fc9c7c9d8e65ee3cfb7abf8e007.

**Description**: As discussed in #529, the change from #535 has broken the AccountEvent list. This PR reverts the change, to make it work again.

**Tested scenarios**: Manual inspection of AccountEvent list.

**Fixed issue**:  #589


AlexandrosMor:
Hello @RickWieman 

Thank you for opening this pr :) Is the AccountEvent removed ? I still she it in [docs](https://docs.adyen.com/api-explorer/#/Account/v6/post/getAccountHolder__resParam_accountHolderStatus-events)

regards,
Alexandros

RickWieman:
Hi @AlexandrosMor,

The way of (de)serialising was changed in v5 (I think), to remove all explicit containers from the JSON. That makes it more intuitive. However, for notifications, you explicitly need to set an `apiVersion`, otherwise they'll fall back to the oldest API specification, which has those containers. That's why @Aleffio introduced the change in #535, but that broke deserialisation of API responses from `v6` (the library's default). We confirmed in #529 that setting the `apiVersion` to `6` on notifications returns the account events in a compatible format, therefore we should revert the change from @Aleffio to be fully compatible with `v6`.

Aleffio (approved):

AlexandrosMor (approved):

cyattilakiss (approved):

Aleffio:
@cyattilakiss @AlexandrosMor let's specify in the next release that "older" notifications will not work anymore, and to create new ones merchants need to specify the `apiVersion` to be `6` in the corresponding api call.
There are also other 'container' fields, which need to be removed as well, will create a ticket for that.

PR #639, opened 2021-10-18
removed details field as mentioned in release notes

wboereboom:
**Description**
removed details as stated [here](https://docs.adyen.com/online-payments/release-notes?integration_type=api#releaseNote=2021-10-06-checkout-api-68)


vpavic:
This resulted in a breaking change upgrading from `17.1.0` to `17.2.0`.

Note that 08b8ec9 removed property named `details` while 6a89a05 that was supposed to reintroduce it in deprecated from actually added property named `inputDetails`.

PR #640, opened 2021-10-19
Release 17.2.0

Morerice:
**Description**
Release 17.2.0


Aleffio:
why are there commits from 17.0.0->17.1.0 as well here?

Issue #622, opened 2021-09-22
Payment methods request returning validation error


andma03:
**Java version**: 15.0.2
**Library version**: 17.0.0
**Description**
When using the PaymentMethodsRequest and sending a request with apiVersion 67 one will receive a validation error saying:
'Structure of PaymentMethodsRequest contains the following unknown fields: [enableRealTimeUpdate]'

This field was deprecated but not removed in PaymentMethodsRequest bean.

Using an objectmapper which does not include null values works, but i think deprecated field/s should be removed.

AlexandrosMor:
Hello @andma03,

Thank you for raising this issue. We opened a PR and we are going to merge it as soon as possible.

kind regards,
Alexandros
Adyen

PR #648, opened 2021-11-04
Remove enableRealTimeUpdate

AlexandrosMor:
**Description**

When using the PaymentMethodsRequest and sending a request with apiVersion 67 one will receive a validation error:
'Structure of PaymentMethodsRequest contains the following unknown fields: [enableRealTimeUpdate]'
This field was deprecated but not removed in PaymentMethodsRequest bean.
**Tested scenarios**

**Fixed issue**:  
fixes: #622


Morerice (approved):

wboereboom (approved):
Approving this PR as an intermediate solution. All Checkout types should be regenerated and updated.

Aleffio (approved):

PR #654, opened 2021-12-07
Fix flaky json test

YujiaYan0709:
**Description**
In these tests, when testing the serialization, originally it compares the serialized String with the expected String. But the order of serialized field is non-deterministic and is, therefore, a flaky test. It might fail when the order does not match, but with the correct output.

**Tested scenarios**
Tested when changing the order of fields in JSON object by using Nondex: https://github.com/TestingResearchIllinois/NonDex

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,11 @@
     </build>
     <dependencies>
         <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.0</version>
+        </dependency>
+        <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
             <version>1.15</version>
--- a/src/test/java/com/adyen/CheckoutTest.java
+++ b/src/test/java/com/adyen/CheckoutTest.java
@@ -124,6 +124,8 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import org.json.JSONException;
+import org.skyscreamer.jsonassert.JSONAssert;
 
 /**
  * Tests for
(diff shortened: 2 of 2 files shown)


michaelpaul:
These tests look deterministic enough to me. I ran them multiple times without failures. They don't seem to rely on external or random factors that would cause flakiness.

YujiaYan0709:
Thanks for the quick response. I understand that flaky tests may only fail in a very rare case. 

According to https://www.json.org/json-en.html and https://www-cs.stanford.edu/~anjiang/papers/ZhangETAL21DexFix.pdf, the serialization of Java objects into JSON strings can produce JSON strings with different order of fields. But the JSON
specification does not specify the order of fields. Comparing the JSON String directly using `assertEquals` assumes to match the order as well and violates the unordered assumption. 

Even the test doesn't fail right now, it could fail in the future when the java version upgrades or runs under a different environment. It's still a potential issue in the program. Using `jsonassert` to compare can ultimately avoid this issue and make your program more robust. These fixes do not change or break the test but only improve it. 

I can fix the similar assert issues in this file later, and let you decide if you want to accept this fix.

YujiaYan0709:
To add more details, `Gson.toJson()` uses `getDeclaredFields()` to serialize class fields, but the order of fields returned is non-deterministic (see https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields) and the specification claimed that "the elements in the returned array are not sorted and are not in any particular order". Therefore, the test should be based on the assumption that class fields can appear in any order.

michaelpaul:
Hi @YujiaYan0709, thanks for the clarification. I agree that the tests may fail under certain circumstances or in the future. 

The main concern that made us hold back a bit was the introduction of an extra library (even if in test scope only). If your improvement relied only on our current dependencies, it would be much easier to get everyone's approval.

michaelpaul:
This has been resolved in #681, thanks for the contribution.

PR #666, opened 2021-12-23
Add support for encoded SSC

liambell-le:
**Description**
ProtectedCardData does not work to add a SSC. Support for a direct encoded string is required for Secondary Security Code on the Adyen Terminal. This change will add encode the SSC taken from the Card Details, and allow it to be set in the PaymentInstrumentData object.

### DIFF
--- a/src/main/java/com/adyen/model/nexo/ContentInformation.java
+++ b/src/main/java/com/adyen/model/nexo/ContentInformation.java
@@ -1,5 +1,8 @@
 package com.adyen.model.nexo;
 
+import com.adyen.serializer.SSCSerializer;
+import com.google.gson.annotations.JsonAdapter;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
@@ -72,7 +75,11 @@ public class ContentInformation {
      */
     @XmlElement(name = "ContentType", required = true)
     protected ContentType contentType;
-
+    
+    @XmlElement(name = "ssc")
+    @JsonAdapter(SSCSerializer.class)
+    protected SSC ssc;
+    
     /**
      * Gets the value of the envelopedData property.
      *
@@ -179,6 +186,24 @@ public class ContentInformation {
      */
     public void setContentType(ContentType value) {
         this.contentType = value;
+    }
+
+    /**
+     * Gets the value of the ssc property.
+     *
+     * @return possible      object is     {@link SSC }
+     */
+    public SSC getSsc() {
+        return ssc;
+    }
+
+    /**
+     * Sets the value of the ssc property.
+     *
+     * @param value allowed object is     {@link SSC }
+     */
+    public void setSsc(SSC value) {
+        this.ssc = value;
     }
 
 }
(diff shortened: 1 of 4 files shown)


wboereboom:
Hi @liambell-le ,

I've spoken to our POS-experts and they have informed me that ssc is already supported through the `ProtectedCardData` field by Base64-encoding it as an object such as `{ "ssc": "XXXXXXXX" }`.

The proposed new `ProtectedSSCData`-field  is not part of our Nexo API Specification, and it is not implemented on our API.

Can you please confirm the aforementioned method does not work for you? And can you provide us an example request/response so we can further investigate? You can post that either here, or take it up with your Implementation Manager if you do not feel comfortable communicating on this thread.

I hope your issue will be resolved shortly!

Kind Regards,
Wouter
Adyen

chadhav-ae:
> Hi @liambell-le ,
> 
> I've spoken to our POS-experts and they have informed me that ssc is already supported through the `ProtectedCardData` field by Base64-encoding it as an object such as `{ "ssc": "XXXXXXXX" }`.
> 
> The proposed new `ProtectedSSCData`-field is not part of our Nexo API Specification, and it is not implemented on our API.
> 
> Can you please confirm the aforementioned method does not work for you? And can you provide us an example request/response so we can further investigate? You can post that either here, or take it up with your Implementation Manager if you do not feel comfortable communicating on this thread.
> 
> I hope your issue will be resolved shortly!
> 
> Kind Regards, Wouter Adyen

Hi @wboereboom ,

Here is the request we are trying to build. Please suggest an alternate for it.

image

wboereboom:
Hi @liambell-le,

This request looks good to me.
You do not need any additions to the library to send this request. Encoding the ssc and placing it in the `ProtectedCardData` should be done in your own implementation. The library can then be used to perform the terminal request.

chadhav-ae:
> Hi @liambell-le @chadhav-ae,
> 
> This request looks good to me. You do not need any additions to the library to send this request. Encoding the ssc and placing it in the `ProtectedCardData` should be done in your own implementation. The library can then be used to perform the terminal request.

Hi @wboereboom , 

We are using this library to set data and send as terminal request(We are not creating our external model classes). If you see the PaymentInstrumentData.java have protectedCardData which we were trying to set the value. But the return of that variable is ContentInformation.java which doesn't have ssc xml element. Kindly please suggest.

wboereboom:
Hi @chadhav-ae,

If that is the case, why not add the ssc field in `ContentInformation.java` so it can be encoded and serialized just like the other fields in that class?

Kind Regards,
Wouter 
Adyen

chadhav-ae:
> Hi @chadhav-ae,
> 
> If that is the case, why not add the ssc field in `ContentInformation.java` so it can be encoded and serialized just like the other fields in that class?
> 
> Kind Regards, Wouter Adyen

Hi @wboereboom, we did tried that too. But the problem we are getting is the encoded data is coming as 
image
which has extra ssc tag. So, this also didn't helped.

wboereboom:
Hi @chadhav-ae,

I understand. We recognize this issue and I am currently consulting our POS-experts to find a suitable solution.
If we cannot fix this on a short term, we might allow for `ProtectedCardData` to be a String, so you could encrypt the data on your end and send it like that.

chadhav-ae:
> Hi @chadhav-ae,
> 
> I understand. We recognize this issue and I am currently consulting our POS-experts to find a suitable solution. If we cannot fix this on a short term, we might allow for `ProtectedCardData` to be a String, so you could encrypt the data on your end and send it like that.

Okay Thanks @wboereboom Let us know what they suggest. Also, once you received positive or negative feedback, can you please add the appropriate code for the fix as well.

wboereboom:
Fixed in #669

PR #669, opened 2021-12-30
change ProtectedCardData type from ContentInformation to String

wboereboom:
**Description**
ContentInformation is not parsed yet at the terminal side. As an intermediate solution ProtectedCardData will be changed to type string so that it may be parsed by the terminal

**Tested scenarios**

**Fixed issue**:  #666


AlexandrosMor (approved):

andrei-dediu (approved):
LGTM !

PR #706, opened 2022-03-17
Run coveralls on pull requests

michaelpaul:
**Description**

Coveralls is not running for external contributors because they cannot push to this repo.


wboereboom (approved):
You might have to edit the github secrets as well!

PR #728, opened 2022-04-22
Adds all missing TransactionStatusEnum values

basstager:
When using the Java library in our application we can't use the Transaction.TransactionStatusEnum in a typed way, because it is currently missing values.
This PR adds all the missing values according to the Docs (https://docs.adyen.com/pt/reporting/settlement-reconciliation/transaction-level/marketplace-payments-accounting-report#record-types).


wboereboom:
Hi @basstager,

When comparing the statuses with our [API specs](https://github.com/Adyen/adyen-openapi/blob/main/json/FundService-v6.json#L941), it seems not all statuses you have proposed are part of our API specification.
Let me know what the case would be for adding these even though they are not part of the specification, and I will inquire as to why they are not part of our API specification.

Kind Regards,
Wouter
Adyen

basstager:
Hi @wboereboom,

We are downloading the Marketplace Payments accounting report from Adyen and then reading the data. The column 'record type' contains values that can't be found anywhere in the Java client in an enum, which is why I want to add them. I'm not sure if they have to be added in the `Transaction.TransactionStatusEnum`, where I added them now, or in `AccountHolderTransactionListRequest.TransactionStatusesEnum`.

I updated the PR to contain all the values from the API spec you linked.

wboereboom:
Hi @basstager,

The answer to your question is maybe both, depending on where. you want to use this Enum. `AccountHolderTransactionListRequest.TransactionStatusesEnum` is only used for the `/accountHolderTransactionList ` endpoint. `Transaction.TransactionStatusEnum` is used in MarketPay Notifications.

We are currently investigating if the enum states you proposed in our initial request need to be added to our specs. If they do, we will add them to the library in a future release.

On top of that we are working on mechanisms to prevent inconsistent models like these in future releases! Please bear with us while we work on these features 🙏 . 

Kind Regards,
Wouter
Adyen

basstager:
@wboereboom Then it seems like it is correct what we are currently doing, which is using the `Transaction.TransactionStatusEnum` and this is where the Enum should be extended.

Any updates on the investigation regarding the enum states?

wboereboom:
Hey @basstager,

No news so far.
In the mean time I can however accept your Merge Request so you can move forwards with your implementation.
Hope it'll be smooth sailing for you from here on onwards 🤞 

Kind Regards,
Wouter
Adyen

basstager:
Hi @wboereboom, just checking in to see if this PR can be merged and released.

wboereboom (approved):
Sorry for the belated response @basstager , looks good to me!

AlexandrosMor (approved):

Issue #457, opened 2020-11-05
"Refund with data" is not supported.


phmp:
**Library version**: 10.0.0
**Description**
I cannot find path “/refundWithData”.
There is eventCode “REFUND_WITH_DATA” but is not used anywhere.
Can you add support for unreferenced refund?

Aleffio:
Hi @phmp ,

Thank you for opening the issue.
We are looking into the possibility to add this endpoint,
I am creating an internal ticket,

Kind Regards,
Alessio
Adyen

SrinivasKG1:
Hi Team, Is there an update on this issue..when refundwithdata will be supported in java-adyen-ali-library

jillingk:
Hi @SrinivasKG1,

Sorry for the super delayed respons. I don't think we will implement this any time soon unfortunately. You're more than welcome to open a PR though! 

Best, Jilling
Adyen

PR #757, opened 2022-06-14
Add refundWithData endpoint

astiskala:
**Description**
Added support for /refundWithData endpoint

**Fixed issue**:  #457


AlexandrosMor:
@astiskala this is an

Issue #707, opened 2022-03-24
[PW-6494] CreateAccountHolderRequest.LegalEntityEnum is missing items


ramisaa:
**Describe the bug**
The LegalEntityEnum seems to be lagging behind. Based on the documentation there are five different items, but only Business and Individual are currently supported. We'd need at least the NonProfit and PublicCompany. The identical LegalEntityEnum from CreateAccountHolderResponse has all the defined items. 

https://docs.adyen.com/platforms/account-holders-and-accounts?tab=create-public-company-request_5#create-an-account-holder

jillingk:
Hi @ramisaa,

Thanks for letting us know, indeed the models were not up to date anymore. I've created a PR to update it, should be approved rather soon! If there is anything else I can help you with let me know.

Best, Jilling
Adyen

PR #769, opened 2022-07-06
[PW-6494] renewed the CreateAccountHolderRequest model

jillingk:
**Description**
LegalEntityEnum missed some values, so I added the new model and ran the tests.

**Fixed issue**:  #707

### DIFF
--- a/src/main/java/com/adyen/model/marketpay/CreateAccountHolderRequest.java
+++ b/src/main/java/com/adyen/model/marketpay/CreateAccountHolderRequest.java
@@ -14,15 +14,20 @@
  *
  * Adyen Java API Library
  *
- * Copyright (c) 2017 Adyen B.V.
+ * Copyright (c) 2022 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  */
 package com.adyen.model.marketpay;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/adyen/model/marketpay/CreateAccountHolderRequest.java
```diff
@@ -32,87 +37,82 @@
...
-    @SerializedName("description")
-    private String description = null;
-
     @SerializedName("accountHolderCode")
     private String accountHolderCode = null;
 
     @SerializedName("accountHolderDetails")
     private AccountHolderDetails accountHolderDetails = null;
 
-    @SerializedName("processingTier")
```
wboereboom, line 47: Do we still need to add this `@SerializedName` annotation? Can you test that?

michaelpaul (approved):

src/main/java/com/adyen/model/marketpay/CreateAccountHolderRequest.java
jillingk: I just tried the unit tests for this endpoint without the annotation and it still passed. I guess that means we can leave it out?

wboereboom (approved):

Issue #775, opened 2022-07-19
Nonce Initial Vector Block Random Function Replacement


ykalayy:
**Java version**: 11
**Library version**: 18.0.0

**Is your feature request related to a problem? Please describe.**
I think It is not a problem. Might be a question. Isn't replacing new Random with singleton SecureRandom much safer?

**Describe the solution you'd like**
Hi, I am new to this repo.
Isn't safer to use singleton SecureRandom instead of new Random for each IV nonce block generation.

**Describe alternatives you've considered**
Even I think the current implementation is safe. Changing nonce initial vector generation with singleton secure random should be much safer whenever especially we use CBC.
**NexoCrypto.java**
```java
  /**
     * Generate a random iv nonce
     */
    private byte[] generateRandomIvNonce() {
        byte[] ivNonce = new byte[NEXO_IV_LENGTH];
        new Random().nextBytes(ivNonce);
        return ivNonce;
    }
```
TO
```java
    /**
     * Generate a random iv nonce
     */
    private byte[] generateRandomIvNonce() {
        byte[] ivNonce = new byte[NEXO_IV_LENGTH];
        SECURE_RANDOM.nextBytes(ivNonce);
        return ivNonce;
    }
```

**Additional context**
None

jillingk:
Hey @ykalayy,

Thanks a lot for contributing! From what I'm reading it seems indeed that simply leveraging Random() for IVnonce might be vulnerable against attacks where adversaries guess this value to break CBC (which would be feasible with random?). Besides the predictability of Random(), it also seems we might also benefit from larger sample group so that we never encounter the same IV twice.

I'm going to pass this onto someone who's more into this stuff since I'm not quite the crypto guru, however your solution looks all good to me. Very good job spotting this one!

Best, Jilling
Adyen

ykalayy:
Hi @jillingk 
Thanks for the response. I keep following the PR. If any objection, I think we can update the PR together

Thanks, Yusuf

PR #776, opened 2022-07-19
Random is replaced with SecureRandom at nonce IV block generation

ykalayy:
**Description**
Random replacement with singleton SecureRandom

**Tested scenarios**
Run UTs

**Fixed issue**:  #775

### DIFF
--- a/src/main/java/com/adyen/terminal/security/NexoCrypto.java
+++ b/src/main/java/com/adyen/terminal/security/NexoCrypto.java
@@ -41,11 +41,13 @@ import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.Random;
+import java.security.SecureRandom;
 
 import static com.adyen.model.terminal.security.NexoDerivedKey.NEXO_IV_LENGTH;
 
 public class NexoCrypto {
+
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
 
     public SaleToPOISecuredMessage encrypt(
             String saleToPoiMessageJson, MessageHeader messageHeader, SecurityKey securityKey) throws Exception {
@@ -149,7 +151,7 @@ public class NexoCrypto {
      */
     private byte[] generateRandomIvNonce() {
         byte[] ivNonce = new byte[NEXO_IV_LENGTH];
-        new Random().nextBytes(ivNonce);
+        SECURE_RANDOM.nextBytes(ivNonce);
         return ivNonce;
     }
 }


src/main/java/com/adyen/terminal/security/NexoCrypto.java
```diff
@@ -41,12 +41,14 @@
...
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.Random;
+import java.security.SecureRandom;
 
 import static com.adyen.model.terminal.security.NexoDerivedKey.NEXO_IV_LENGTH;
 
 public class NexoCrypto {
 
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
jillingk, line 50: Perhaps consider replacing ```SecureRandom()``` with ```SecureRandom.getInstanceStrong()```

ykalayy: @jillingk Very thanks for the suggestion
From what I'm reading SecureRandom.getInstanceStrong() uses dev/random which is blocked in Linux systems. I prefer to use the default one which uses dev/urandom and under the hood the getInstanceStrong is no secure than the default one.

jillingk (approved):

src/main/java/com/adyen/terminal/security/NexoCrypto.java
jillingk: Hi @ykalayy, that makes sense. I will get this PR approved. Thanks again for contributing!

ykalayy:
Could you please review? @AlexandrosMor @wboereboom @Morerice  @michaelpaul

src/main/java/com/adyen/terminal/security/NexoCrypto.java
```diff
@@ -149,7 +151,7 @@ private void validateHmac(byte[] receivedHmac, byte[] decryptedMessage, NexoDeri
      */
     private byte[] generateRandomIvNonce() {
         byte[] ivNonce = new byte[NEXO_IV_LENGTH];
-        new Random().nextBytes(ivNonce);
+        SECURE_RANDOM.nextBytes(ivNonce);
```
jillingk, line 154: Hey @ykalayy,

We just double checked this with our security team, since we're not too comfortable with this kind of crypto stuff ourselves. They recommended us to use ```SecureRandom.getInstance("NativePRNGNonBlocking", "SUN")```, because this would presumably provide the securest PRG without blocking over all platforms. If you want to see all the various options for the SecureRandom class you can go over to this [site](https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#securerandom-number-generation-algorithms). Let me know your thoughts on this!

Best, Jilling
Adyen

AlexandrosMor (approved):

wboereboom:
Resolved in #818

PR #782, opened 2022-08-03
Adding the SSC support

arajska:
**Description**
Adding the support for the SSC code for PW-6825

**Tested scenarios**
Added the additional field as discussed internally. 

**Fixed issue**:  PW-6825

### DIFF
--- a/src/main/java/com/adyen/model/terminal/SaleToAcquirerData.java
+++ b/src/main/java/com/adyen/model/terminal/SaleToAcquirerData.java
@@ -44,6 +44,7 @@ public class SaleToAcquirerData {
     private String tenderOption;
     private Map<String, String> additionalData;
     private String authorisationType;
+    private int ssc;
     private static final Gson PRETTY_PRINT_GSON = new GsonBuilder().setPrettyPrinting().create();
 
     public SaleToAcquirerData() {
@@ -162,6 +163,14 @@ public class SaleToAcquirerData {
         this.authorisationType = authorisationType;
     }
 
+    public int getSsc(){
+        return ssc;
+    }
+
+    public void setSsc(int ssc){
+        this.ssc = ssc;
+    }
+
     public static Gson getPrettyPrintGson() {
         return PRETTY_PRINT_GSON;
     }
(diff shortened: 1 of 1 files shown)


AlexandrosMor (dismissed):

wboereboom (approved):

src/main/java/com/adyen/model/terminal/SaleToAcquirerData.java
```diff
@@ -162,6 +163,14 @@ public void setAuthorisationType(String authorisationType) {
         this.authorisationType = authorisationType;
     }
 
+    public int getSsc(){
```
acampos1916, line 166:
```suggestion
    public int getSsc() {
```

acampos1916 (approved):

wboereboom:
@arajska Unit tests are failing on this. Could it be the case that ssc is always set to  `0`, even if it is not contained in the request/response?

arajska:
@wboereboom I did not set any default value for it. I followed the same logic as for the other fields that are already there - declared it at the top of the class and created a getter and a setter for it.

wboereboom:
Gotcha. I think the issue pops up because an int has value `0` by default, whereas a String is `null`. That's why the test fails about this `0` popping up unexpectedly when parsing SaleToAcquirerData.

Do you know if this could cause any issues?

arajska:
Interesting that you're bumping into this issue. I was able to make payments with it no problem though I admit that I did not try to set it to 0. I don't have a terminal at hand at the moment so can't check it for sure but I feel like it should work - I also got a successful transaction while having the ssc as a string so I don't think the terminal cares. But the gift card provider might care about it. Will need to double check this with the LPM team.

wboereboom:
The issue would be  when trying a payment without a sec value. The value would then still be set to 0, but we would have to test this.
---
Wouter Boereboom
Integration Tools Developer

contact: adyen.com/contact <https://www.adyen.com/contact>

Adyen Headquarters
Simon Carmiggeltstraat 6-50
1011 DJ Amsterdam, The Netherlands

	 <https://www.adyen.com/>

PR #802, opened 2022-09-05
PW-6294: Add OpenAPI model templates

michaelpaul:
**Description**

Templates retrieved with:

```
openapi-generator author template -g java -o templates --library okhttp-gson
```

OpenAPI Generator version: 6.0.1


wboereboom (dismissed):

templates/libraries/okhttp-gson/pojo.mustache
```diff
@@ -599,23 +597,4 @@ public class {{classname}} {{#parent}}extends {{{.}}} {{/parent}}{{#vendorExtens
   }
 {{/hasChildren}}
 
- /**
```
wboereboom: Why not include these methods? Might be of use for some devs

michaelpaul: Because they depend on this other object `JSON` (`import {{invokerPackage}}.JSON`) which is not a model and only works for a single service at a time. We can investigate how to integrate it to the library on another ticket.

wboereboom (approved):

AlexandrosMor (approved):

PR #834, opened 2022-10-06
Pw 7050/update templates

wboereboom:
**Description**
Update templates so they are properly able to parse oneOf models based on allowed Enum Values

**Tested scenarios**
Tested using Checkout models and unit tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,6 +231,21 @@
             <artifactId>swagger-annotations</artifactId>
             <version>2.2.2</version>
         </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <version>2.1.1</version>
+        </dependency>
+        <dependency>
+            <groupId>io.gsonfire</groupId>
+            <artifactId>gson-fire</artifactId>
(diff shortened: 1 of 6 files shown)


AlexandrosMor (approved):

AlexandrosMor (dismissed):

pom.xml
```diff
@@ -231,6 +231,21 @@
...
+            <artifactId>javax.ws.rs-api</artifactId>
+            <version>2.1.1</version>
+        </dependency>
+        <dependency>
+            <groupId>io.gsonfire</groupId>
+            <artifactId>gson-fire</artifactId>
+            <version>1.8.5</version>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
```
michaelpaul, line 245: This adds a new HTTP client to the library, can we remove this dependency via templates? Otherwise, it's better to migrate `--library` to `apache-httpclient` to match our current client.

michaelpaul (changes requested):

pom.xml
wboereboom: I can probably narrow it down to the classes the library models use from this package.

michaelpaul (approved):

PR #850, opened 2022-10-18
Release 18.1.2

wboereboom:
**Description**
- #818 
- #848

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,11 @@ $(services): target/spec $(openapi-generator-jar)
 		--skip-validate-spec \
 		--model-package $(subst /,.,com.adyen.model.$@) \
 		--library $(library) \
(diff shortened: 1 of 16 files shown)


michaelpaul (approved):

Makefile
```diff
@@ -5,7 +5,7 @@ openapi-generator-cli:=java -jar $(openapi-generator-jar)
 
 generator:=java
 library:=okhttp-gson
-services:=payments
+services:=payments checkout
```
AlexandrosMor, line 8: Do we want to add the checkout or specific service in the `makefile` ?

wboereboom: I don't see why not. It does not influence this release, and we'll be using it going forwards

AlexandrosMor (approved):

PR #853, opened 2022-10-19
[PW-7223] Fix for serializing Date objects to ISO 8601 format for Checkout

jillingk:
**Description**
Added a custom adapter for date objects in Checkout classes for merchant requesting this. Also added Unit Tests that we can keep and extend when we add with the new solution for generated models, as it's nice to check the Date object to ISO8601 conversion during these upgrades. 

**Tested scenarios**
Unit Tests in DateSerializationTest.java


wboereboom (approved):

AlexandrosMor (approved):

src/main/java/com/adyen/model/checkout/PaymentSessionRequest.java
```diff
@@ -159,7 +160,7 @@ public ChannelEnum read(final JsonReader jsonReader) throws IOException {
     private String countryCode = null;
 
     @SerializedName("dateOfBirth")
-    @JsonAdapter(DateSerializer.class)
+    @JsonAdapter(DateTimeISO8601Serializer.class)
     private Date dateOfBirth = null;
```
michaelpaul, line 164: Time is not necessary in this field. I think we can stick to `DateSerializer` here. 

https://docs.adyen.com/api-explorer/Checkout/69/post/paymentSession#request-dateOfBirth

jillingk: Actually I thought we could make it easier for ourselves and make this one ISO8601 as well since the format entails YYYY-MM-DD format and filling the field in as such will result in the same format. Plus I tried sending a full ISO8601 string (with timezone) on Postman and the API just accepts that as well, so I feel like everything ISO8601 is the way to go? Let me know what you think :)

michaelpaul (approved):

PR #854, opened 2022-10-20
[PW-7256] Add Bank and Issuing Fund Transfers V3 API

AlexandrosMor:
**Description**

Added the models and the following endpoints:

/transactions
/transactions/{id}
/transfers

**Tested scenarios**

Added unit tests based on the open api examples


src/test/java/com/adyen/TransfersTest.java
```diff
@@ -0,0 +1,45 @@
+package com.adyen;
+
+import com.adyen.model.transfer.*;
+import com.adyen.service.Transactions;
+import com.adyen.service.Transfers;
```
michaelpaul, line 5: I think that for a small service like this we don't need two classes, how about combining all endpoints into a single class?

wboereboom: Ola! Devil's advocate here 😈 !

@AlexandrosMor  is actually following our API documentation [here](https://docs.adyen.com/api-explorer/transfers/3/overview)

Do we follow suit or deviate?

src/main/java/com/adyen/service/Resource.java
```diff
@@ -55,11 +55,12 @@ public Resource(Service service, String endpoint, List<String> requiredFields) {
      * Request using json String
      *
      * @param json request json
+     * @param post
      * @return request
      * @throws ApiException ApiException
      * @throws IOException  IOException
      */
-    public String request(String json) throws ApiException, IOException {
+    public String request(String json, ApiConstants.HttpMethod post) throws ApiException, IOException {
```
michaelpaul, line 63: Unused variable `post`. Why do we need to change all the existing services now?

michaelpaul (changes requested):

src/test/java/com/adyen/TransfersTest.java
michaelpaul: It's a bit overkill, but we can have two small classes. Still, they should be combined via single parent class. We don't have two separate `com.adyen.service.Transactions ` and `com.adyen.service.Transfers` services, it's only one **Transfers API**.

PR #856, opened 2022-10-21
[PW-7256] Add Bank and Issuing Fund Transfers V3 API

AlexandrosMor:
**Description**

Added the models and the following endpoints:

/transactions
/transactions/{id}
/transfers

**Tested scenarios**

Added unit tests based on the open api examples


michaelpaul (changes requested):
See my previous comment: https://github.com/Adyen/adyen-java-api-library/pull/854#discussion_r1001530658

michaelpaul (approved):

wboereboom (approved):

src/main/java/com/adyen/service/Transfers.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public Transfer transfers(TransferInfo transferInfo) throws IOException, ApiException {
+        String jsonRequest = transferInfo.toJson();
+        TransfersResource transfersResource = new TransfersResource(this, "/transfers");
+        String jsonResult = transfersResource.request(jsonRequest);
+        return Transfer.fromJson(jsonResult);
+    }
+
+    public TransactionSearchResponse list() throws IOException, ApiException {
```
wboereboom, line 35: Maybe it is good to rename these. These endpoints are about transactions, whereas the first one is about a Tranasfer.

wboereboom:
MakeFile and Readme should also get updated

src/main/java/com/adyen/service/Transfers.java
wboereboom: @AlexandrosMor can you look into this?

PR #888, opened 2022-11-15
Update models

AdyenAutomationBot:
OpenAPI spec or templates produced new models on 02-02-2023   by [commit](https://github.com/Adyen/adyen-openapi/commit/1b2b017a6df5da1678b933227ab51c1e66e2821d).


src/main/java/com/adyen/model/balanceplatform/AccountHolderCapability.java
```diff
@@ -121,7 +120,7 @@ public AllowedLevelEnum read(final JsonReader jsonReader) throws IOException {
 
   public static final String SERIALIZED_NAME_PROBLEMS = "problems";
   @SerializedName(SERIALIZED_NAME_PROBLEMS)
-  private List<CapabilityProblem> problems = null;
+  private List<Object> problems = null;
```
michaelpaul, line 123: The `CapabilityProblem` model is in LEM now, we need to review it's usage across services.

michaelpaul (changes requested):

src/main/java/com/adyen/model/balanceplatform/AccountHolderCapability.java
wboereboom: Noted. We will add this as a known caviat ✅

michaelpaul (approved):

AlexandrosMor (approved):

PR #908, opened 2022-11-24
move common models to specific packages

wboereboom:
**Description**
Move all models that are shared between services to their respective packages

**Tested scenarios**
Tested through unit tests


AlexandrosMor (approved):

michaelpaul (commented):
I wonder if we should remove these shared models or embrace them via `--import-mappings Amount=com.adyen.model.Amount`.

wboereboom:
The thing is that if we ever start to autogenerate these library models we'll overwrite them anyway. so figured it was best to get rid of them now.

I think it's better to have duplicates than to deal with the added complexity of shared models between APIs.

michaelpaul (approved):

PR #909, opened 2022-11-24
Added requestOptions to storeDetailAndSubmitThirdParty in Payout API

luggi961:
**Description**
Created new storeDetailAndSubmitThirdParty method with RequestOptions parameter to be able to set idempotencyKey.

### DIFF
--- a/src/main/java/com/adyen/service/Payout.java
+++ b/src/main/java/com/adyen/service/Payout.java
@@ -72,11 +72,7 @@ public class Payout extends Service {
      * @throws ApiException ApiException
      */
     public StoreDetailAndSubmitResponse storeDetailAndSubmitThirdParty(StoreDetailAndSubmitRequest request) throws IOException, ApiException {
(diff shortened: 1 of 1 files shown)


jillingk:
Hi @luggi961,

Thanks a lot for reaching out to us and contributing! The PR itself seems good to me, however one small question; why do you need Idempotency key for this specific payout endpoint? I think we only really use/allow Idempotency keys for the checkout and payments endpoints, hence my question :)

Best, Jilling
Adyen

luggi961:
Hi @jillingk 

Thanks for your message. In the past we had the problem that due to an issue the same payout request was sent twice to Adyen. Therefore, we would like to add this safety mechanism. 

 I checked with our implementation manager from Adyen and he confirmed that idempotency is supported for the Payout API as well.

Best,
Lukas

wboereboom:
Hey @luggi961 ,

Looks good to me :)!

Kind Regards,
Wouter
Adyen

PR #910, opened 2022-11-25
[PW-7449] Remove Parent Class from LEM/BalancePlatform

jillingk:
**Description**
Parent Class is not needed here since we have packages. Removed them, added JSON constructors and modified the UT's


jillingk:
Again don't mind the commits, I merged the BalancePlatform branch herein hence the many commits from that branch.

src/main/java/com/adyen/service/legalentitymanagement/Documents.java
```diff
@@ -4,6 +4,7 @@
 import com.adyen.Service;
 import com.adyen.constants.ApiConstants;
 import com.adyen.model.legalEntityManagement.Document;
+import com.adyen.model.legalEntityManagement.JSON;
```
michaelpaul, line 7: Could you also please rename the models package to `com.adyen.model.legalentitymanagement`?

michaelpaul (commented):
The main part looks good! I just want to include another change here.

src/main/java/com/adyen/service/legalentitymanagement/Documents.java
jillingk: done!

AlexandrosMor (approved):

wboereboom (approved):

PR #911, opened 2022-11-25
Added RREQ data structure into NotificationRequestItem

sabra:
**Description**
Added data structure for Raw 3DS RREQ webhook

**Tested scenarios**
Tested handling of incoming RREQ webhook

### DIFF
--- a/src/main/java/com/adyen/model/Rreq.java
+++ b/src/main/java/com/adyen/model/Rreq.java
@@ -1,0 +1,225 @@
+package com.adyen.model;
+
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+import static com.adyen.util.Util.toIndentedString;
+
+
+/**
+ * RREQ
+ */
+public class Rreq {
+
+    @SerializedName("acsTransID")
+    private String acsTransID;
+
+    @SerializedName("authenticationType")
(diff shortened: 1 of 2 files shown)


sabra:
Hey gyus, could you perhaps look into this a bit more proactively? It takes unnecessary long time. Your coveralls script failed on 
`Failed to execute goal org.eluder.coveralls:coveralls-maven-plugin:4.3.0:report (default-cli) on project adyen-java-api-library: Build error: Either repository token or service with job id must be defined -> [Help 1]`

jillingk:
Hi @sabra,

Apologies for the not so timely response, I guess this one flew a bit under our radar for some reason.

I just looked at your PR and I have a couple of notes. Firstly I cannot actually find the notification event code you are trying to add anywhere 😅 , any chance you have a docs link or screenshot of the notification in question? The second thing is, that at the moment we prefer to not add any models to the notificationRequestItem class but keep it as generic as possible in form of a key-value pair map. This, because we currently are not able to generate any consistent models and as such rather keep it like generic to avoid having outdated models in this library. Hopefully in the future we can actually generate it and have it fully up to date.

If you can show me an example of the webhook that'd be great though, so we can add the event code and you'll at least be able to handle the notification without the library throwing an error. 

Best, Jilling
Adyen

PR #918, opened 2022-12-09
[PW-7442] Fix Byte Array (de)serialization 

jillingk:
**Description**
Ran the model generation to have everything up to date. Also added UT's to test the (de)serialization of the byte-arrays adapter in the JSON files based on these tests in the actual swagger api;
https://github.com/swagger-api/swagger-codegen/pull/7473/files#diff-310cb3de8284569421b36a781c998f7d765ca8d9a6c9261d2db8ba4ffd7dd343

Also added two UT's to test the JSON base64 encoding/decoding of the byte-array fields (where I used an online base64 encoder/decoder to create the validation strings).


src/test/java/com/adyen/PaymentTest.java
```diff
@@ -477,16 +459,80 @@ protected void assertRefused(PaymentResult paymentResult) {
...
+
+        threeDSecureData.setCavv("AQIDBAUGBwgJCgsMDQ4PEBESExQ".getBytes(StandardCharsets.UTF_8));
+        paymentRequest.mpiData(threeDSecureData);
         
         payment.authorise(paymentRequest);
-        
-        String expected = "\"mpiData\":{\"cavv\":\"AQIDBAUGBwgJCgsMDQ4PEBESExQ=\"}";
+
+        // In the json the string is encoded in base64
+        String expected = "\"mpiData\":{\"cavv\":\"QVFJREJBVUdCd2dKQ2dzTURRNFBFQkVTRXhR\"}";
```
michaelpaul, line 515: The previous JSON string expectation is correct: 

```
String expected = "\"mpiData\":{\"cavv\":\"AQIDBAUGBwgJCgsMDQ4PEBESExQ=\"}";
```

The string "AQIDBAUGBwgJCgsMDQ4PEBESExQ=" is already base64 encoded. The string "QVFJREJBVUdCd2dKQ2dzTURRNFBFQkVTRXhR" is another base64 encode (double encoding) of what the user set via:

```
paymentRequest.mpiData(new ThreeDSecureData().cavv("AQIDBAUGBwgJCgsMDQ4PEBESExQ=".getBytes()));
```  

This is happening now because the new [ByteArrayAdapter](https://github.com/Adyen/adyen-java-api-library/blob/1814a41bf7f7509a1a588b3c8d0d921a01b7c32e/src/main/java/com/adyen/model/payments/JSON.java#L222) applies base64 by default assuming the input is not encoded already.  

See also my previous [investigation resources](https://github.com/Adyen/adyen-java-api-library/pull/848#issue-1411156955).

michaelpaul (changes requested):

wboereboom:
This PR got contaminated with a lot of model changes somehow.

jillingk:
> This PR got contaminated with a lot of model changes somehow.

Actually thing is that I had to generate all the models again because I changed the templates in order to make the bytearray not serialise in base64 encoding (since we are assuming that all bytarrays given by merchants are already encoded). Just to double check, @michaelpaul is this how you imagined it?

src/main/java/com/adyen/model/balanceplatform/AccountHolderCapability.java
```diff
@@ -121,7 +120,7 @@ public AllowedLevelEnum read(final JsonReader jsonReader) throws IOException {
 
   public static final String SERIALIZED_NAME_PROBLEMS = "problems";
   @SerializedName(SERIALIZED_NAME_PROBLEMS)
-  private List<CapabilityProblem> problems = null;
+  private List<Object> problems = null;
```
michaelpaul, line 123: The `CapabilityProblem` model has been deleted? A `List<Object>` is not very useful. Can you double check this please?

src/main/java/com/adyen/model/legalentitymanagement/ZoneOffset.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.Map.Entry;
+import java.util.Set;
+
+import com.adyen.model.legalentitymanagement.JSON;
+
+/**
+ * ZoneOffset
+ */
+
+public class ZoneOffset {
```
michaelpaul, line 50: I think this class might be incorrect, it looks like the same issue as PW-7624.

Makefile
```diff
@@ -1,11 +1,11 @@
-openapi-generator-version:=6.0.1
+openapi-generator-version:=6.2.1
```
michaelpaul, line 1: Can we do this upgrade in a separate PR? Ideally, model changes should come out from our bot in a PR like #888. In this scope of this ticket, only `JSON` classes should change.

PR #923, opened 2022-12-23
[PW-7442] Change JSON.java template to fix byte-array adapter

jillingk:
**Description**
Removed base64 encoding from bytearray adapter, and generated the JSON.java classes again.

**Tested scenarios**
UT's


src/test/java/com/adyen/PaymentTest.java
```diff
@@ -475,18 +456,36 @@ protected void assertAuthorised(PaymentResult paymentResult) {
...
         Client client = createMockClientFromFile("mocks/authorise-success.json");
         Payment payment = new Payment(client);
         PaymentRequest paymentRequest = new PaymentRequest();
         paymentRequest.mpiData(new ThreeDSecureData().cavv("AQIDBAUGBwgJCgsMDQ4PEBESExQ=".getBytes()));
         
         payment.authorise(paymentRequest);
-        
-        String expected = "\"mpiData\":{\"cavv\":\"AQIDBAUGBwgJCgsMDQ4PEBESExQ=\"}";
+        // Unicode representation of the string
+        String expected = "\"mpiData\":{\"cavv\":\"AQIDBAUGBwgJCgsMDQ4PEBESExQ\\u003d\"}";
```
wboereboom, line 488: Does the API expect this format?

jillingk: This is unicode, which apparently most APIs would be able to handle just fine. @michaelpaul do you think our API's handle this? I can alternatively remove this by ignoring the html escaping in the gsonbuilder.

jillingk: Just verified with actual request in api logs, API does properly parse this string with the html escaping like that!

michaelpaul: Here it's our previous encoder: https://github.com/Adyen/adyen-java-api-library/blob/311eea70f6515477abe1d9a29db091ab15e2a025/src/main/java/com/adyen/serializer/ByteArrayToStringAdapter.java#L40

I think we can proceed with this new one since its very similar.

michaelpaul (approved):

wboereboom (approved):

PR #924, opened 2022-12-28
PW-6983: Add missing fields from AccountHolderVerificationNotificationContent

michaelpaul:
**Description**

Add missing fields, specially `signatoryCode`, from AFP notification model on v18.


wboereboom (approved):
Should we automate model generation for MarketPay?

PR #943, opened 2023-01-26
[PW-7716] Removing unused depedencies

jillingk:
**Description**
Removed unused dependencies and removed some Jackson specific annotations from MarketPay endpoints.

**Tested scenarios**
UT's


jillingk:
> @jillingk Do we need this [dependency](https://github.com/Adyen/adyen-java-api-library/blob/3f59a2868f2fbd3bc4c61a21598bdbfbc59695c3/pom.xml#L236) ?

So yeah no I don't think we really do. It's this nullable annotation we do in some generated parts of the library, but it's only really for IDE's to recognise nullpointerexceptions or smth? Either way I removed it (through the templates) and everything still worked fine. But I did want to discuss first before I removed it in this PR.
```java
@javax.annotation.Nonnull
```

jillingk:
Okio is for the ByteString type in the ByteArrayAdapter in the *.JSON classes:
```java
@Override
        public byte[] read(JsonReader in) throws IOException {
            switch (in.peek()) {
                case NULL:
                    in.nextNull();
                    return null;
                default:
                    String bytesAsBase64 = in.nextString();
                    ByteString byteString = ByteString.decodeBase64(bytesAsBase64);
                    return byteString.toByteArray();
            }
        }
```
It's generated w the templates, but if we can find an (java native) alternative I'm sure we can change it.

src/main/java/com/adyen/model/marketpay/AccountEvent.java
```diff
@@ -52,7 +51,7 @@ public enum EventEnum {
         @SerializedName("RefundNotPaidOutTransfers")
         REFUNDNOTPAIDOUTTRANSFERS("RefundNotPaidOutTransfers");
 
-        @JsonValue
+
```
wboereboom, line 54: can we not have this blank line?

wboereboom (commented):
Have we made sure removing these did not break any enum (de)serialization?

AlexandrosMor (dismissed):

michaelpaul:
> Okio is for the ByteString type in the ByteArrayAdapter in the *.JSON classes:
> 
> ```java
> @Override
>         public byte[] read(JsonReader in) throws IOException {
>             switch (in.peek()) {
>                 case NULL:
>                     in.nextNull();
>                     return null;
>                 default:
>                     String bytesAsBase64 = in.nextString();
>                     ByteString byteString = ByteString.decodeBase64(bytesAsBase64);
>                     return byteString.toByteArray();
>             }
>         }
> ```
> 
> It's generated w the templates, but if we can find an (java native) alternative I'm sure we can change it.

Let's take a look into https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html.

jillingk:
> > Okio is for the ByteString type in the ByteArrayAdapter in the *.JSON classes:
> > ```java
> > @Override
> >         public byte[] read(JsonReader in) throws IOException {
> >             switch (in.peek()) {
> >                 case NULL:
> >                     in.nextNull();
> >                     return null;
> >                 default:
> >                     String bytesAsBase64 = in.nextString();
> >                     ByteString byteString = ByteString.decodeBase64(bytesAsBase64);
> >                     return byteString.toByteArray();
> >             }
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's generated w the templates, but if we can find an (java native) alternative I'm sure we can change it.
> 
> Let's take a look into https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html.

I actually ended up using the apache library, that we already used, to do this encoding (only one line of code compared to some other solutions). Also moved the okio to test scope, and generated the JSON.java files so it now does not use okio anymore.

pom.xml
```diff
@@ -210,31 +200,20 @@
...
         <dependency>
             <groupId>io.swagger.core.v3</groupId>
             <artifactId>swagger-annotations</artifactId>
             <version>2.2.8</version>
         </dependency>
         <dependency>
-            <groupId>javax.ws.rs</groupId>
-            <artifactId>javax.ws.rs-api</artifactId>
-            <version>2.1.1</version>
+            <groupId>jakarta.ws.rs</groupId>
```
wboereboom, line 214: lets look into this first before just changing it. Are we sure we need to do this?

wboereboom (commented):
do we test base64 (de)serialisation in unit tests?

jillingk:
> do we test base64 (de)serialisation in unit tests?

Yep https://github.com/Adyen/adyen-java-api-library/blob/bd9cd6338aa9fdffb745b5dc20e442503465e54e/src/test/java/com/adyen/PaymentTest.java#L460

pom.xml
jillingk: It's actually just the same package but under a [new name](https://mvnrepository.com/artifact/javax.ws.rs/javax.ws.rs-api).

pom.xml
```diff
@@ -210,31 +200,20 @@
             <version>4.11.0</version>
             <scope>test</scope>
```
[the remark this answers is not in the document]
jillingk: So outside of test, we would use scope>compile (default) for the other dependencies right?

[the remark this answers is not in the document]
wboereboom: Yes!

wboereboom (dismissed):

wboereboom (approved):

michaelpaul (approved):

PR #950, opened 2023-01-30
[PW-7716-extended] Removing javax-annotation dependency

jillingk:
**Description**
```java
<dependency>
    <groupId>com.google.code.findbugs</groupId>
    <artifactId>jsr305</artifactId>
    <version>3.0.0</version>
</dependency>
```
Removed this javax.annotation as in my opinion we should not have it in this library. [Reading here](https://stackoverflow.com/questions/13484202/how-to-use-nullable-and-nonnull-annotations-more-effectively) it seems that this overal does not add much, plus it might even be a bit of a bad habit since merchants should do there own null checks?

**Tested scenarios**
UT


michaelpaul (approved):

michaelpaul:
We can generate explicit `null` checks later if necessary.

wboereboom (approved):

PR #953, opened 2023-01-31
New template

antolo-arch:
**Description**
New service template for management which should work without the plugin and also add overloading the method.
One method is with all possible parameters (required and optional) and the other one is with required ones only


wboereboom (approved):
regenerated Checkout locally. all seems to work very well. Good Job

jillingk (approved):

Issue #988, opened 2023-03-23
createSweep function in BalanceAccounts service is returning wrong type


sharipsg:
**Describe the bug**
In com.adyen.service.balanceplatform.BalanceAccounts -> **createSweep** function is returning wrong type `BalanceSweepConfigurationsResponse`. It is supposed to return `SweepConfigurationV2`, similar to **retrieveSweep** function.
openapi ref - https://github.com/Adyen/adyen-openapi/blob/main/json/BalancePlatformService-v2.json#L930
API doc ref - https://docs.adyen.com/api-explorer/balanceplatform/latest/post/balanceAccounts/_balanceAccountId_/sweeps#responses-200

**How to reproduce**
Create a sweep for a balance account using SDK and an exception is thrown when deserializing because of wrong type casting

**Expected behavior**

**Screenshots**
<img width="1008" alt="image" src="https://user-images.githubusercontent.com/106537957/227113483-306982ec-d384-46be-ae4f-741ecf9758d8.png">

**Desktop (please complete the following information):**
 - Library Version: 19.0.0

**Additional context**
Also a new `category` field is returned in the create sweep API response
```
{
    "id": "<ID>",
    "schedule": {
        "type": "cron",
        "cronExpression": "0 * * * *"
    },
    "status": "active",
    "targetAmount": {
        "currency": "USD",
        "value": 0
    },
    "triggerAmount": {
        "currency": "USD",
        "value": 0
    },
    "type": "push",
    "category": "bank",
    "counterparty": {
        "transferInstrumentId": "<ID>"
    },
    "currency": "USD"
}
```

sharipsg:
This is fixed in v20.0.0

PR #989, opened 2023-03-27
Bugfix For BCL v2 Endpoints and Add Missing Endpoint

jillingk:
**Description**
This PR fixes a wrong return type for createSweep method in BalanceAccounts.java. It also adds a missing endpoint and regenerates BCL v2.

**Tested scenarios**
UT

**Fixes**
#988


src/main/java/com/adyen/service/balanceplatform/BalanceAccounts.java
```diff
@@ -36,11 +36,11 @@ public BalanceSweepConfigurationsResponse listSweeps(String balanceAccountId, Ma
         return BalanceSweepConfigurationsResponse.fromJson(jsonResult);
     }
 
-    public BalanceSweepConfigurationsResponse createSweep(String balanceAccountId, SweepConfigurationV2 request) throws IOException, ApiException {
+    public SweepConfigurationV2 createSweep(String balanceAccountId, SweepConfigurationV2 request) throws IOException, ApiException {
```
wboereboom, line 39: Will we have to modify this service again when this object name gets fixed?

jillingk: Nah I think this is the actual name, as that class has been around for a while.

wboereboom (approved):

AlexandrosMor (approved):

PR #990, opened 2023-03-29
fix: static initialisers to avoid side effects in the JSON constructors

sp00m:
As the title says :)

### DIFF
--- a/src/main/java/com/adyen/model/balanceplatform/JSON.java
+++ b/src/main/java/com/adyen/model/balanceplatform/JSON.java
@@ -85,7 +85,7 @@ public class JSON {
         return clazz;
     }
 
-    {
+    static {
         GsonBuilder gsonBuilder = createGson();
         gsonBuilder.registerTypeAdapter(Date.class, dateTypeAdapter);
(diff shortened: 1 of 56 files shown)


wboereboom:
Hey @sp00m ,
Thanks a lot for contributing to our repository.
Could you please elaborate on the side effects you are trying to avoid?

Kind Regards,
Wouter
Adyen

sp00m:
Hey @wboereboom,

The previous code relied on an [instance initializer block](https://www.baeldung.com/java-static-instance-initializer-blocks#instance-initializer-block) to set a static variable, which is a side-effect: as the block executes each time a new instance is created, each new instance of `JSON` will _overwrite_ the static `gson` instance. That's also the reason why you had to have `new JSON();` statements scattered a bit everywhere I suppose.

Using a [static block](https://www.baeldung.com/java-static-instance-initializer-blocks#static-block) instead will get rid of this oddity and ensure `gson` is set only once: when the class first loads.

Hope that makes sense. It's rather minor for sure, but I just came across it and the fix was rather quick, so I thought why not contributing :)

jillingk (approved):

PR #1001, opened 2023-04-18
ITT-141 Automating Services in Java Lib

jillingk:
**Description**
Created this draft PR to show progress on the java service automation.
TODO: 
- baseURL formatting (test/live) -> done
- Overloading methods for non required params -> done
- Create single file service -> done
- Method description -> done


wboereboom:
Should we remove the existing service? Or is it better to deprecate first?

jillingk:
> Should we remove the existing service? Or is it better to deprecate first?

@wboereboom difficult to decide tbh. Would be easier to migrate if we deprecate it, but then again we're cluttering the lib quite a lot..

Makefile
```diff
@@ -52,6 +59,62 @@ $(services): target/spec $(openapi-generator-jar)
...
+		-g $(generator) \
+		-t templates \
+		-o $(output) \
+		--reserved-words-mappings configuration=configuration \
+		--ignore-file-override ./.openapi-generator-ignore \
+		--skip-validate-spec \
+		--model-package $(subst /,.,com.adyen.model.$@) \
+		--library $(library) \
+		--api-package com.adyen.service.$@ \
+        --api-name-suffix Service \
```
michaelpaul, line 82: I kind like the `Api` suffix, it produces classes like `PaymentsApi` and `ModificationsApi`. Checkout as a whole is a service and these classes are parts of the whole, but not a service on their own. They also form the main API offered by the library to use our REST API's, seems like a good choice of word. ;) 

```suggestion
        --api-name-suffix Api \
```

Makefile
```diff
@@ -52,6 +59,62 @@ $(services): target/spec $(openapi-generator-jar)
 	mv $(output)/$(models)/$@ $(models)/$@
 	mv $(output)/$(models)/JSON.java $(models)/$@
 
+# Full service + models automation
+bigServices:=balanceplatform checkout storedValue payments payout management legalentitymanagement transfers
+singleFileServices:=balancecontrol binlookup dataprotection storedvalue posterminalmanagement recurring
+
+all: $(bigServices) $(singleFileServices)
```
michaelpaul, line 66: The `all` target is invoked when you run `make` (without any target). I would rather to have two targets, `models` for when generating only models and `services` for both models and service classes.

```suggestion
services: $(bigServices) $(singleFileServices)
```

michaelpaul (approved):
Great job @jillingk! 🚀

michaelpaul:
The way query string parameters are passed is changing from `Map<String, String>` to actual method arguments. The order of path, body, and query has also changed. 

This change is good and can be done to existing services (e.g. Management and Balance Platform) in a new major release.

Makefile
```diff
@@ -52,6 +59,62 @@ $(services): target/spec $(openapi-generator-jar)
...
+		-g $(generator) \
+		-t templates \
+		-o $(output) \
+		--reserved-words-mappings configuration=configuration \
+		--ignore-file-override ./.openapi-generator-ignore \
+		--skip-validate-spec \
+		--model-package $(subst /,.,com.adyen.model.$@) \
+		--library $(library) \
+		--api-package com.adyen.service.$@ \
+        --api-name-suffix Api \
```
wboereboom, line 82:
```suggestion
		--api-name-suffix Api \
```

Makefile
```diff
@@ -52,6 +59,62 @@ $(services): target/spec $(openapi-generator-jar)
...
+		-c templates/libraries/okhttp-gson/config.yaml \
+		-o $(output) \
+		--reserved-words-mappings configuration=configuration \
+		--ignore-file-override ./.openapi-generator-ignore \
+		--skip-validate-spec \
+		--model-package $(subst /,.,com.adyen.model.$@) \
+		--library $(library) \
+		--additional-properties customApi=$@ \
+		--api-package com.adyen.service \
+        --api-name-suffix Api \
```
wboereboom, line 107:
```suggestion
		--api-name-suffix Api \
```

src/main/java/com/adyen/Service.java
```diff
@@ -49,4 +50,27 @@ public void setApiKeyRequired(boolean apiKeyRequired) {
         isApiKeyRequired = apiKeyRequired;
     }
 
+    public String createBaseURL(String url) {
```
jillingk, line 53:
```suggestion
    protected String createBaseURL(String url) {
```

michaelpaul (approved):

antolo-arch (approved):

AlexandrosMor (approved):

PR #1002, opened 2023-04-19
Put back JsonProperty as it's now impossible to deserialize the json object

Kearny:
**Description**
Put back the @JsonProperty name to be able to deserialize the object from JSON

**Tested scenarios**
With and without the annotation

**Fixed issue**:


wboereboom:
hey @Kearny ,
thanks for raising this PR.
We  removed Jackson as a dependency and are now relying on Gson for (de)serialization.
We also have a [notificationHandler utility class](https://github.com/Adyen/adyen-java-api-library/blob/develop/src/main/java/com/adyen/notification/NotificationHandler.java) to help support this process.

Please let us know if you still experience these serialisation issues using Gson.

Kind Regards,
Wouter
Adyen

Issue #1037, opened 2023-05-23
SaleToPOIResponses for Twint in Switzerland can't be deserialized


scriptease:
**Describe the bug**
SaleToPOIResponses for Twint in Switzerland can't be deserialized

**How to reproduce**
Decode a SaleToPOIResponses from Twint in Switzerland using
`TerminalAPIGsonBuilder.create().fromJson(json, TerminalAPIResponse.class);`

**Expected behavior**
The result contains the SaleToAcquirerData and doesn't complain about an object being expected instead of the string.

**Desktop (please complete the following information):**
 - Java Version: 1.8
 - Library Version: 17.1.0, 20.0.0

**Additional context**
Pull request will follow

scriptease:
PR: https://github.com/Adyen/adyen-java-api-library/pull/1038

jillingk:
Hi @scriptease,

We've just released your fix! Thanks again for contributing, it's much appreciated :)

Best, Jilling
Adyen

scriptease:
Awesome 🙏

PR #1038, opened 2023-05-23
Add SaleToAcquirerDataSerializer deserialisation

scriptease:
**Description**
Adds missing deserialization of SaleToAcquirerData for SaleToPOIResponses, e.g. when using the payment provider Twint in Switzerland

**Tested scenarios**
* Base 64 Encodeded SaleToAcquirerData was parsed correctly. 

**Fixed issue**:  #[1037](https://github.com/Adyen/adyen-java-api-library/issues/1037)

### DIFF
--- a/src/main/java/com/adyen/model/terminal/SaleToAcquirerData.java
+++ b/src/main/java/com/adyen/model/terminal/SaleToAcquirerData.java
@@ -20,13 +20,18 @@
  */
 package com.adyen.model.terminal;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.commons.codec.binary.Base64;
+
 import com.adyen.model.applicationinfo.ApplicationInfo;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import org.apache.commons.codec.binary.Base64;
-
-import java.util.Map;
-import java.util.Objects;
 
 public class SaleToAcquirerData {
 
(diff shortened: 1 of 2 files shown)


jillingk:
Hi @scriptease,

Thanks a lot for taking the time to contribute! Quick question, what kind of request did Twint send to get a SaleToAcquirerData object back in the response? I'm asking because I haven't seen this field in a response before so I just wanted to double check we actually do need to deserialize this.

Your PR looks all good though! Maybe a unit test would be nice, but I can help you with that as well.

Best, Jilling
Adyen

scriptease:
It happened in live/production so I don't think I can share the whole response here, however the base64 and the decoded version is just meta data with a applicationInfo.


```
      "SaleData" : {
        "SaleTransactionID" : {
          "TransactionID" : "...",
          "TimeStamp" : "2023-04-15T14:13:07.000Z"
        },
        "SaleToAcquirerData" : "ewogICJhcHBsaWNhdGlvbkluZm8iOiB7CiAgICAiYWR5ZW5MaWJyYXJ5IjogewogICAgICAibmFtZSI6ICJhZHllbi1qYXZhLWFwaS1saWJyYXJ5IiwKICAgICAgInZlcnNpb24iOiAiMTcuMC4wIgogICAgfQogIH0KfQ=="
      },
```


```
{
  "applicationInfo": {
    "adyenLibrary": {
      "name": "adyen-java-api-library",
      "version": "17.0.0"
    }
  }
}

```

scriptease:
>  I just wanted to double check we actually do need to deserialize this.

We could also ignore it using @ Expose while parsing the json, but without the fix it throws a json processing exception which prohibited us from processing payments, so we already created our own bugfix release and deployed it.

I created the ticket & pr mostly for fellow developers that might run into a similar edge case, I never heard about twint investigating the issue.

jillingk:
Okay makes sense to me, I'll approve it. I think in the past we would not encounter these responses, hence why we didn't have this yet. Thanks again for the quick PR!

jillingk (approved):

wboereboom (approved):

PR #1041, opened 2023-05-24
Update models

AdyenAutomationBot:
OpenAPI spec or templates produced new models on 24-07-2023   by [commit](https://github.com/Adyen/adyen-openapi/commit/a0dff9e1aad25f30897d7171997365bf7979e6d8).


src/test/java/com/adyen/PaymentTest.java
```diff
@@ -468,7 +477,7 @@ public void TestByteArrayToJSONString() throws Exception {
         
         payment.authorise(paymentRequest);
         // Unicode representation of the string
-        String expected = "\"mpiData\":{\"cavv\":\"AQIDBAUGBwgJCgsMDQ4PEBESExQ\\u003d\"}";
+        String expected = "\"mpiData\":{\"cavv\":\"QVFJREJBVUdCd2dKQ2dzTURRNFBFQkVTRXhRPQ==\"}";
```
michaelpaul, line 480: This is expectation is not correct. It's based on double base64 encode.

michaelpaul (changes requested):

src/test/java/com/adyen/PaymentTest.java
jillingk: Right, good catch! Completely forgot about that issue.. Will fix it.

jillingk:
@michaelpaul Fixed the byteArray!

src/main/java/com/adyen/model/checkout/SubMerchant2.java
```diff
@@ -0,0 +1,254 @@
...
+@JsonPropertyOrder({
+  SubMerchant2.JSON_PROPERTY_ADDRESS,
+  SubMerchant2.JSON_PROPERTY_ID,
+  SubMerchant2.JSON_PROPERTY_MCC,
+  SubMerchant2.JSON_PROPERTY_NAME,
+  SubMerchant2.JSON_PROPERTY_TAX_ID
+})
+@JsonTypeName("SubMerchant-2")
+
+public class SubMerchant2 {
```
Aleffio, line 43: is this another "2" case? should this be fixed on the specs first?

wboereboom: This is an unfortunate case. I'm wondering whether it's worth it to go back and fix this right now.

jillingk: The submerchant2 class has a ID parameter and Address instead of city + country param. I propose we refactor the submerchant2 class to SubMerchantInfo manually in this library, and then change it later on main to SubMerchantInfo so we don't have any breaking changes.

wboereboom (approved):

michaelpaul (approved):

PR #1043, opened 2023-06-01
Update models no checkout

wboereboom:
**Description**
Copy of the update/models PR without updates to checkout due to current openApi spec challenges


jillingk (approved):

michaelpaul (approved):
It still contains a `src/main/java/com/adyen/model/balanceplatform/JSONObject.java`, but I guess we will fix it later.

wboereboom:
It is part of the error object. We cannot get rid of that one yet unfortunately

PR #1048, opened 2023-06-06
[ITT-578] Switch from GSON serializer to Jackson

jillingk:
**Description**
Wrote templates to switch to jackson deserialization. Had to add custom toJson() and fromJson() methods so it we didn't have to move around the layout of the service classes. Also had to make objectMapper static which we have to discuss. Lastly I'm still working on the import we're using and verifying all the oneOf serialization works.

Basic TODO:
- Stricter OneOf deserialization based on type enum -> done
- Loose validation for additional and non-required parameters -> done
- Numeric parameters are nullable (capture delay hour debacle) -> done
- Readonly not giving problems -> done
- DateSerialization (both serialization/deserialization are iso8601 without adjusting to context time zone) -> done

Notes:

- Base64 encoding is enforced. Will throw error if the byte[] values are not proper base64 during serialization (seems like its not a problem as it will have to be encoded anyways right?)
- If oneOf models deserialize against more than one model I added a warning (had to allow it because of super similar models in bcl).

Migrate rest of services:
- Banking webhooks -> done
- Payment (General) webhooks ->
- Remove all Classic platforms remnants (hop/fund/account/webhooks) -> done, but we should really deprecate it first in this major release cycle
- Terminal API -> not worth the trouble migrating, so I guess we're keeping some gson dependencies for now?


pom.xml
```diff
@@ -181,6 +181,16 @@
         </plugins>
     </build>
     <dependencies>
+        <dependency>
```
wboereboom, line 184: Should we also remove Gson from POM?

wboereboom:
Maybe it would be a good idea to see if all (d)serialization tests and adapters will still work if we make this switch. Looks like a good first draft 👍

PR #1051, opened 2023-06-08
[ITT-572] Process webhooks with Optional field

jillingk:
**Description**
Implemented @michaelpaul's feedback regarding the webhooks. This way merchants get use the Optional fields to check the webhook type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,8 @@ boolean authenticity = hmacValidator.validateHMAC(hmacKey, signKey, payload);
 ~~~~
 If this bool returns true, one can proceed to deserialize against the desired webhook type:
 ~~~~ java
-BankingWebhookHandler webhookHandler = new BankingWebhookHandler();
-AccountHolderNotificationRequest accountHolderNotificationRequest = webhookHandler.getAccountHolderNotificationRequest(payload);
+BankingWebhookHandler webhookHandler = new BankingWebhookHandler(payload);
+AccountHolderNotificationRequest accountHolderNotificationRequest = webhookHandler.getAccountHolderNotificationRequest().get();
 ~~~~
 ### Proxy configuration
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -176,8 +176,8 @@ boolean authenticity = hmacValidator.validateHMAC(hmacKey, signKey, payload);
 ~~~~
 If this bool returns true, one can proceed to deserialize against the desired webhook type:
 ~~~~ java
-BankingWebhookHandler webhookHandler = new BankingWebhookHandler();
-AccountHolderNotificationRequest accountHolderNotificationRequest = webhookHandler.getAccountHolderNotificationRequest(payload);
+BankingWebhookHandler webhookHandler = new BankingWebhookHandler(payload);
+AccountHolderNotificationRequest accountHolderNotificationRequest = webhookHandler.getAccountHolderNotificationRequest().get();
```
michaelpaul, line 180: The `Optional.get` method is used when you are sure about the `payload` contents. Normally that won't be the case, so a user can attach a callback for each case and only the valid one will be invoked:  

```suggestion
// onAccountHolderNotificationRequest
webhookHandler.getAccountHolderNotificationRequest().ifPresent((AccountHolderNotificationRequest event) -> {
    System.out.println(event.getData().getBalancePlatform());
});
// onBalanceAccountNotificationRequest
webhookHandler.getBalanceAccountNotificationRequest().ifPresent((BalanceAccountNotificationRequest event) -> {
    System.out.println(event.getData().getBalanceAccount());
});

```

michaelpaul (changes requested):
Looking good! Just some small tweaks.

michaelpaul (approved):

wboereboom (approved):

PR #1054, opened 2023-06-10
ITT-598: Strict OneOf validation for deserialization against multiple models

jillingk:
This PR fixes the sweepschedule issue. Check the SweepConfigurationV2Schedule.java class to see how the deserialization has been changed in order to account for the issue.


AlexandrosMor (approved):

michaelpaul (dismissed):

michaelpaul (approved):
We are making `oneOf` validation strict to avoid ambiguities and allowing other (simple) models to ignore (new) fields missing from the specifications at the time the library was built. 

New fields might eventually appear in responses received by users of old versions. These fields will be ignored by the library instead of crashing. When a user needs the ignored field, one must upgrade to a new library version.

wboereboom (approved):

antolo-arch (approved):

PR #1072, opened 2023-06-29
move capital API into transfers

wboereboom:
**Description**
rename Transfers service to TransfersApi and move capital endpoints into TransfersApi.
This PR also renames the function names on TransfersApi to be more expressive. underlying functionality remains unchanged

**Tested scenarios**
Updated unit tests


Makefile
```diff
@@ -67,8 +65,8 @@ $(modelGen): target/spec $(openapi-generator-jar)
 	mv $(output)/$(models)/JSON.java $(models)/$@
 
 # Full service + models automation
-bigServices:=balanceplatform checkout payout management legalentitymanagement transfers
-singleFileServices:=balancecontrol binlookup dataprotection storedvalue posterminalmanagement recurring payment capital
+bigServices:=balanceplatform checkout payout management legalentitymanagement
+singleFileServices:=balancecontrol binlookup dataprotection storedvalue posterminalmanagement recurring payment transfers
```
antolo-arch, line 69: Why is transfers in the single file services? does it not have 3 sub parts, transfers, transactions and capital?

wboereboom: I did that because it has less than 10 endpoints. I thought it would be a bit much to split it in several services with less than 3 methods each

jillingk: I think we might as well keep to the structure defined in the api explorer. Also could you regenerate the models again @wboereboom? The current ones still use GSON.

jillingk (approved):

antolo-arch (approved):

PR #1075, opened 2023-07-03
[ITT-578] Remove Classic Platforms and lingering GSON depedencies

jillingk:
**Description**
This PR removes the classic platform endpoints. It also moves the GSON dependency from the Service class so spring applications do not have to import this dependency by default


antolo-arch (approved):

src/main/java/com/adyen/service/CheckoutUtility.java
```diff
@@ -27,12 +27,15 @@
...
 import com.adyen.service.resource.checkoututility.OriginKeys;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
 
 import java.io.IOException;
 
 public class CheckoutUtility extends ApiKeyAuthenticatedService {
 
+    protected static final Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
```
wboereboom, line 38: Should this be changed to Jackson?

wboereboom (approved):

antolo-arch (approved):

PR #1093, opened 2023-07-25
Automation/models Fix

jillingk:
**Description**
This PR fixes the issues with the automation/models branch. Inline schema mapping needed to be adjusted again and model naming becomes simpler.


wboereboom (approved):

michaelpaul (approved):
It's also worth checking if the readme samples are still correct.

PR #1094, opened 2023-07-26
Add services workflow

AlexandrosMor:
**Description**
Add the services workflow for java

**Tested scenarios**
Generate the models and the services local


.github/workflows/service.yml
```diff
@@ -0,0 +1,31 @@
...
+          token: ${{ secrets.ADYEN_AUTOMATION_BOT_ACCESS_TOKEN }}
+          committer: ${{ secrets.ADYEN_AUTOMATION_BOT_EMAIL }}
+          author: ${{ secrets.ADYEN_AUTOMATION_BOT_EMAIL }}
+          base: develop
+          branch: automation/services
+          title: ${{ steps.vars.outputs.pr_title }}
+          body: ${{ steps.vars.outputs.pr_body }}
+          add-paths: |
+            Adyen/Model
+            Adyen/Service
```
michaelpaul, line 31:
```suggestion
            src/main/java/com/adyen/model
            src/main/java/com/adyen/service
```

michaelpaul (changes requested):

.github/workflows/service.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          token: ${{ secrets.ADYEN_AUTOMATION_BOT_ACCESS_TOKEN }}
+          committer: ${{ secrets.ADYEN_AUTOMATION_BOT_EMAIL }}
+          author: ${{ secrets.ADYEN_AUTOMATION_BOT_EMAIL }}
+          base: develop
+          branch: automation/services
+          title: ${{ steps.vars.outputs.pr_title }}
+          body: ${{ steps.vars.outputs.pr_body }}
+          add-paths: |
+            src/main/java/com/adyen/model
```
Aleffio, line 30: just to understand, if this script is for both /service and /model, do we plan to remove the `models.yml` file?

Aleffio (approved):

michaelpaul (changes requested):
Let's keep the `models.yml` workflow and introduce a separate `services.yml`. Models change more frequently and maintain compatibility, which is the opposite of how services evolve normally.

michaelpaul (approved):

jillingk (approved):

PR #1112, opened 2023-08-18
Allow passing baseURL directly when constructing API services

sp00m:
The implementation of `createBaseURL(...)` forces us to use HTTPS, but we need our app servers to use HTTP and let our service mesh catch and encrypt the outgoing requests before they leave our system (transparent from your point of view, the requests will always reach you through HTTPS).

One backward-compatible approach could be as proposed here: allow the `baseURL` to be passed to the constructor directly. It would greatly simplify our usage, given we currently have to hack around by overriding `createBaseURL(...)` thanks to its `protected` visibility.

Let me know what you think.

### DIFF
--- a/templates/libraries/jersey3/api.mustache
+++ b/templates/libraries/jersey3/api.mustache
@@ -16,11 +16,19 @@ import java.util.Map;
 
 {{#operations}}
 public class {{classname}} extends Service {
+
+    public static final String API_VERSION = "{{{version}}}";
+
     protected String baseURL;
 
     public {{classname}}(Client client) {
         super(client);
         this.baseURL = createBaseURL("{{{basePath}}}");
+    }
+
+    public {{classname}}(Client client, String baseURL) {
+        super(client);
+        this.baseURL = baseURL;
     }
 {{#operation}}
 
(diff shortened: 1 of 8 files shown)


templates/libraries/jersey3/api.mustache
```diff
@@ -16,12 +16,20 @@ import java.util.Map;
 
 {{#operations}}
 public class {{classname}} extends Service {
+
+    public static final String API_VERSION = "{{{version}}}";
```
sp00m, line 20: Takes the value of [`#/info/version`](https://swagger.io/specification/#info-object).

wboereboom: Do you need this to construct your baseUrl in a maintainable fashion?

wboereboom:
Hey @sp00m,

Thanks for contributing. We did not consider this scenario in our design.
LGTM!

Kind Regards,
Wouter
Adyen

wboereboom (approved):

templates/libraries/jersey3/api.mustache
sp00m: Yes exactly, I suppose your SDK knows which version of the API it supports, so having it available here would allow us rely on it instead of hardcoding it ourselves.

jillingk:
Hey @sp00m, looks good to me as well. Could use with some method documentation explaining the use of this overload, however I can also pick that up. Cheers!

templates/libraries/jersey3/api.mustache
wboereboom: Yes, we use API specs which are tightly coupled to the actual API. 👍

PR #1114, opened 2023-08-22
Cleaning unused config properties

sp00m:
We started to get a little bit confused by the several properties we could configure when instantiating a client, and after a deeper look, it seems that some of them aren't used (anymore?), so I hope you don't mind me proposing this modest PR where I removed the properties from `com.adyen.Config` whose getters were used nowhere.

Pretty minor though, feel free to close it if you prefer keeping them for now.


wboereboom:
Hi @sp00m,

Thank you for contributing once more!
I'm sorry to hear the configurations have been confusing. We have recently made a push to make our service classes more self-contained, and it seems our configuration classes have not entirely caught up to that fact yet!
I'm going to accept your contribution, and we'll look into the client and config in the near future, to see what else we should remove.

Kind Regards,
Wouter
Adyen

wboereboom (approved):

PR #1162, opened 2023-10-24
Replace apache.commons Base64 with jdk Base64

AlexandrosMor:
**Description**

`java.util.Base64` is more performant than `org.apache.commons.codec.binary.Base64`.
We already use `java.util.Base64` in some of the Unit tests, but we still use `org.apache.commons.codec.binary.Base64` in the code.
Replace the apache commons Base64 with the ones from the jdk.
**Tested scenarios**

All tests are passing


templates/libraries/okhttp-gson/JSON.mustache
```diff
@@ -218,7 +217,7 @@ public class JSON {
                     return null;
                 default:
                     String bytesAsBase64 = in.nextString();
-                    return Base64.decodeBase64(bytesAsBase64);
+                    return Base64.getDecoder().decode(bytesAsBase64);
```
jillingk, line 220:
```suggestion
                    return Base64.getDecoder().decode(bytesAsBase64);
```
This is actually not the proper mustache template dir. You need the other one :)

jillingk: -> jersey3

AlexandrosMor: There is not base64 related code in `templates/libraries/jersey3/`, right ? @jillingk

src/main/java/com/adyen/util/HMACValidator.java
```diff
@@ -22,7 +22,7 @@
 
 import com.adyen.model.notification.Amount;
 import com.adyen.model.notification.NotificationRequestItem;
-import org.apache.commons.codec.binary.Base64;
+import java.util.Base64;
 import org.apache.commons.codec.binary.Hex;
```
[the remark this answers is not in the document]
AlexandrosMor: @Aleffio I think this outside of this ticket scope, right ?

Aleffio (approved):

jillingk (approved):

PR #1186, opened 2023-11-28
Refactored some of the Smells

DrashtiNavadiya:
**Description**

**Tested scenarios**

**Fixed issue**:


wboereboom:
Hi @DrashtiNavadiya,

Thanks for proposing these changes.
Unfortunately most of them will be overwritten by our automation, so they are not feasible in this form.
I also see a .jar file in your PR that I do not recognise, and there is no description or input that warrants adding it. 
I am going to close this PR to prevent the risk of changes which that will be overwritten and unknown files being added to our repository.

Kind Regards,
Wouter
Adyen

PR #1190, opened 2023-11-29
Jersey3 OpenAPI templates: add attribute description as comment

gcatanese:
**Description**
I'd like to suggest a change to the Jersey3 templates so that each pojo attribute includes the OpenAPI description as a comment, helping developers to understand the property and how to use it, for example:

```
  // The 8-digit bank account number, without separators or whitespace.
  private String accountNumber;
```


templates/libraries/jersey3/pojo.mustache
```diff
@@ -82,6 +82,10 @@ public class {{classname}} {{#parent}}extends {{{.}}} {{/parent}}{{#vendorExtens
   private {{{datatypeWithEnum}}} {{name}}{{#required}}{{#defaultValue}} = {{{.}}}{{/defaultValue}}{{/required}}{{^required}} = null{{/required}};
   {{/isContainer}}
   {{^isContainer}}
+  {{#description}}
+
+  // {{{.}}}
```
jillingk, line 87:
```suggestion
/**
* {{{.}}}
*/
```
How about we use actual java descriptions so ideas will pick up the description when devs hover over it?

jillingk: Could also add this atop the constructors/setters if we'd like to

michaelpaul (approved):

jillingk (approved):

PR #1219, opened 2024-01-02
Generated tapi

jillingk:
**Description**
Generated Tapi models.


README.md
```diff
@@ -285,7 +285,7 @@ import com.adyen.Client;
 import com.adyen.enums.Environment;
 import com.adyen.service.TerminalCloudAPI;
 import com.adyen.model.nexo.*;
-import com.adyen.model.terminal.*;
+import com.adyen.model.terminalOld.*;
```
wboereboom, line 288: what is terminalOld?

src/main/java/com/adyen/model/terminal/AmountsResp.java
```diff
@@ -0,0 +1,324 @@
...
+   * Get authorizedAmount
+   * minimum: 0.0
+   * maximum: 99999999.999999
+   * @return authorizedAmount
+  **/
+  @ApiModelProperty(required = true, value = "")
+  @JsonProperty(JSON_PROPERTY_AUTHORIZED_AMOUNT)
+  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
+
+  public BigDecimal getAuthorizedAmount() {
```
wboereboom, line 110: This type seems fishy. isn't amount always an integer or long?

src/main/java/com/adyen/terminal/TerminalAPIRequest.java
```diff
@@ -0,0 +1,124 @@
+/*
```
wboereboom, line 1: let's get rid of these headers now

jillingk: done

src/main/java/com/adyen/model/terminal/AmountsResp.java
jillingk: Actually we already used this type in the old nexo models, let me double check tho

README.md
```diff
@@ -284,7 +284,7 @@ In order to submit In-Person requests with [Terminal API over Cloud](https://doc
 import com.adyen.Client;
 import com.adyen.enums.Environment;
 import com.adyen.service.TerminalCloudAPI;
-import com.adyen.model.nexo.*;
+import com.adyen.terminal.*;
```
wboereboom, line 287: Should we make it more clear what we should import from where? I think the `com.adyen.terminal.*` import is pretty fuzzy compared to how other services work

jillingk: Updated the imports in the readme, should we proceed with this?

jillingk: Added implicit imports

wboereboom (approved):

AlexandrosMor (approved):

Issue #1265, opened 2024-03-26
Typo in TaxReportingClassification.BusinessTypeEnum


zzeng2:
**Describe the bug**
In TaxReportingClassification.BusinessTypeEnum:
FINANCIALINSTITUTION_("financialInstitution.");
this value has extra underscore in name and extra dot in value, which caused exception when using fromValue.

**How to reproduce**
Call BusinessTypeEnum.fromValue("financialInstitution");

**Expected behavior**
The value should be correct string.

**Screenshots**

**Desktop (please complete the following information):**
 - Java Version: [19.0.2-zulu]
 - Library Version: [24.0.0]

**Additional context**
none

DjoykeAbyah:
Hi @zzeng2! 

Thank you for reaching out to us. The latest release addressed this issue. We recommend updating to the latest version to resolve it!

Djoyke
Adyen

PR #1266, opened 2024-03-27
TaxReportingClassification fix for issue #1265

tddouglas:
**Description**
Removed unnecessary underscore and period from TaxReportingClassification type

**Tested scenarios**
 Called BusinessTypeEnum.fromValue("financialInstitution"); to verify there was no error. 

**Fixed issue**:  #1265


wboereboom:
Hi @tddouglas ,

This code is automated, hence this change would be overwritten if we were to merge it. 
It does however look like this is the result of a bug in our openApi specifications.

I will make sure to reach out to the appropriate stakeholders so this will get fixed.

Kind Regards,
Wouter
Adyen

Issue #1308, opened 2024-06-11
AccountHolderNotificationRequest cant parse AccountHolderNotificationRequest webhook


mattbachmann-toast:
**Describe the bug**
I received the following request in my test environment

```
{
    "data": {
        "balancePlatform": "Toast-US-Issuing",
        "accountHolder": {
            "contactDetails": {},
            "description": "Bachmann test holder",
            "legalEntityId": "LE32CKZ223225Q5KT9J773SQC",
            "reference": "I think we can put restaurantGuid here",
            "capabilities": {
                "receiveFromBalanceAccount": {
                    "enabled": "true",
                    "requested": "true",
                    "allowed": "false",
                    "verificationStatus": "pending"
                },
                "sendToBalanceAccount": {
                    "enabled": "true",
                    "requested": "true",
                    "allowed": "false",
                    "verificationStatus": "pending"
                },
                "sendToTransferInstrument": {
                    "enabled": "true",
                    "requested": "true",
                    "allowed": "false",
                    "verificationStatus": "pending"
                },
                "issueCard": {
                    "enabled": "true",
                    "requested": "true",
                    "allowed": "false",
                    "verificationStatus": "pending"
                },
                "useCard": {
                    "enabled": "true",
                    "requested": "true",
                    "allowed": "false",
                    "verificationStatus": "pending"
                }
            },
            "id": "AH32957223227M5KT9JQWFHWB",
            "status": "Active"
        }
    },
    "environment": "test",
    "type": "balancePlatform.accountHolder.created"
}
```

For the life of me I was not able to get this parsed.  I tried

```
BankingWebhookHandler(body).accountHolderNotificationRequest.ifPresent { 
            ....redacted...
        }
```

But `BankingWebhookHandler(body).accountHolderNotificationRequest` was always null. So I tried to process it a bit more manually.

```
AccountHolderNotificationRequest.fromJson(body)
```

This crashes with 

```
Cannot construct instance of `com.adyen.model.configurationwebhooks.AccountHolder$StatusEnum`, problem: Unexpected value 'Active'
```

This is because the enum does not have `Active` defined. It has `active`. This is when I figured it was a bug and here I am.

**How to reproduce**
Pass the provided webhook body to `AccountHolderNotificationRequest.fromJson(body)`

**Expected behavior**
Seems like it should parse

**Screenshots**

[...]

PR #1309, opened 2024-06-12
Enums can come back in different cases then specifically documented, …

mattbachmann-toast:
…so handle enums in a case insensitive manner

This is my attempt at addressing https://github.com/Adyen/adyen-java-api-library/issues/1308  but there are other solutions for sure.

But I leave it to yall to decide on if this can be merged and what would need to happen for that.


mattbachmann-toast:
I did run the regen and that did work. But I did not include those changes in this PR since I figured thats a separate process

PR #1361, opened 2024-09-26
Log parsing error

gcatanese:
**Description**
When the BankingWebhookHandler serialization fails, the exception is suppressed and not reported. This doesn't provide the developers with any visible feedback or message.

This PR logs the error message to help developers understand what happens.

**Tested scenarios**
Using the latest Adyen Java library with Banking webhook `v1`, the models/enums no longer match, and the BankingWebhookHandler returns an empty Optional instance.


jillingk (approved):
Nice one!

PR #1367, opened 2024-10-08
Improve deprecation information adding deprecatedInVersion and deprecatedMessage in the Javadoc

gcatanese:
The library deprecates correctly the attributes/endpoints that are marked as deprecated in the OpenAPI spec files.  

This PR improves the existing code by adding the additional information (`deprecatedInVersion`, `deprecatedMessage`) in the Javadoc.
Formatting has also been improved removing unnecessary extra lines after getter/setter methods.

PR #1370, opened 2024-10-14
Skip escaping descriptions in mustache templates

gcatanese:
**Description**
Escaping class/attribute description is breaking the Java API library pipeline as the generated javadoc contains the character `>`, which is not allowed and should be escaped.

This PR revert partially the changes introduced in PR #1367


gcatanese:
Superseded by #1371

PR #1404, opened 2024-12-10
fix(exception): better handle non-JSON response bodies

sp00m:
When an error occurs and the response body isn't JSON (e.g. a misconfiguration of the base path triggering an HTML error page as a response), the code swallows the original error and instead throws a parsing exception.

This PR ensures parsing exceptions are ignored so that the original issue remains available.


gcatanese:
@sp00m Thank you for the contribution. It is a useful change, but the unit test fails. Can you maybe have a look? Thank you

sp00m:
> @sp00m Thank you for the contribution. It is a useful change, but the unit test fails. Can you maybe have a look? Thank you

The unit tests do pass, it's your SonarCloud pipeline that seems to fail because somehow misconfigured?

gcatanese (approved):

PR #1407, opened 2024-12-17
Adjust Region class with valid regions plus endpoint mappings for terminal api live

DjoykeAbyah:
Adjusted Region class to define region constants and their associated Terminal API endpoint mappings.

Added a RegionTest class to validate:

The list of valid regions (VALID_REGIONS) matches expected values.
The endpoint mappings (TERMINAL_API_ENDPOINTS_MAPPING) are consistent with the provided regions.
Adjusted the Client class to include public static endpoint constants that the Region class references.


src/main/java/com/adyen/Client.java
```diff
@@ -13,6 +13,10 @@ public class Client {
     public static final String LIB_VERSION = "32.0.0";
     public static final String TERMINAL_API_ENDPOINT_TEST = "https://terminal-api-test.adyen.com";
     public static final String TERMINAL_API_ENDPOINT_LIVE = "https://terminal-api-live.adyen.com";
```
Kwok-he-Chu, line 15: Let's keep endpoint variable names consistent with those above

src/main/java/com/adyen/Client.java
```diff
@@ -13,6 +13,10 @@ public class Client {
...
+    public static final String ENDPOINT_TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
```
Kwok-he-Chu, line 16: `"https://terminal-api-live.adyen.com"` has been added twice (see `TERMINAL_API_ENDPOINT_LIVE`)

Kwok-he-Chu: Perhaps moving these endpoints-names out of the Client into the respective terminal-api service or its own config would be better, what do you think?

DjoykeAbyah: Yes that makes sense. I'll remove the duplicate. about moving the endpoint names to a separate config. I'm not sure I see that for .Net "https://terminal-api-live.adyen.com" is added as CloudApiEndPointEULive in the clientconfig. https://github.com/Adyen/adyen-dotnet-api-library/blob/af0f3b7259736e961992840e7b9514befdcc6789/Adyen/Constants/ClientConfig.cs#L9C28-L9C52 but for php it is like you suggested for Java that EU is the default live endpoint https://github.com/Adyen/adyen-php-api-library/blob/af262a702228c473bbb4c00f9d7000f0d85ad7e5/src/Adyen/Client.php#L30C41-L30C80. Do you suggest creating a separate terminalconfig for all the libraries and set a specific ENDPOINT_TERMINAL_CLOUD_EU_LIVE?

src/test/java/com/adyen/RegionTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class RegionTest {
+    @Test
+    public void testValidRegions() {
+
+        // Convert Regions.VALID_REGIONS to lowercase strings for comparison
+        List<String> actual = Regions.VALID_REGIONS.stream()
+        .map(region -> region.name().toLowerCase())
```
Kwok-he-Chu, line 18: nit: add indenting when using map and collect

src/test/java/com/adyen/RegionTest.java
```diff
@@ -0,0 +1,64 @@
...
+            actual
+        );
+    }
+
+    @Test
+    public void testTerminalApiEndpointsMapping() {
+
+        // Convert TERMINAL_API_ENDPOINTS_MAPPING keys to lowercase strings for comparison
+        Map<String, String> actual = Regions.TERMINAL_API_ENDPOINTS_MAPPING.entrySet()
+        .stream()
```
Kwok-he-Chu, line 43: nit: add indenting when using stream and collect

src/test/java/com/adyen/ClientTest.java
```diff
@@ -38,6 +48,83 @@ public void testConfigLiveClient() {
...
+        Client liveClient = new Client(liveConfig);
+        
+        // call the method and assert the result
+        String result = liveClient.getCloudEndpoint(region, liveConfig.getEnvironment());
+        Assert.assertEquals("Unexpected endpoint for region: " + region, expectedEndpoint, result);
+    }
+
+    @Test
+    public void testRegionsNotInMappingThrowException() {
+        List<Region> unmappedRegions = new ArrayList<>();
```
Kwok-he-Chu, line 98: Since India is currently the only region that is not in the map, we can rewrite `testRegionsNotInMappingThrowException` and assert that "In"<->mapping should throw an exception, that will keep simplify this test

Kwok-he-Chu (commented):
Nice work, just some minor nitpicks

src/main/java/com/adyen/enums/Region.java
```diff
@@ -1,8 +0,0 @@
-package com.adyen.enums;
-
-public enum Region {
```
Kwok-he-Chu, line 3: Can we keep the original `Region` enum class, otherwise this change will be breaking

src/test/java/com/adyen/ClientTest.java
DjoykeAbyah: I was thinking about the possibility more regions would be added without a region url in the future but it makes more sense to now only test for "In"

src/main/java/com/adyen/Client.java
```diff
@@ -139,6 +144,25 @@ public void setEnvironment(Environment environment, String liveEndpointUrlPrefix
         }
     }
 
+    /**
+     * @param region The region for which the endpoint is requested. If null or the region is not found, defaults to default EU endpoint.
+     */
+    public String getCloudEndpoint(Region region, Environment environment) {
```
Kwok-he-Chu, line 150: The `getCloudEndpoint` method is not used when the `Client` is instantiated. See `setEnvironment(..)` function

Kwok-he-Chu (changes requested):

src/main/java/com/adyen/Client.java
DjoykeAbyah: I'm not sure I understand you correctly. The `getCloudEndpoint` is an optional way to determine your regional endpoint. If we were to adjust `setEnvironment` we have to add the region parameter to the method.

src/main/java/com/adyen/Client.java
Kwok-he-Chu: Keep the current strings as is, and default to EU

src/main/java/com/adyen/Client.java
```diff
@@ -127,16 +131,36 @@ public void setEnvironment(Environment environment) {
...
+                return Region.TERMINAL_API_ENDPOINTS_MAPPING.get(Region.EU);
+            }
+            if (!Region.TERMINAL_API_ENDPOINTS_MAPPING.containsKey(region)) {
+                throw new IllegalArgumentException("TerminalAPI endpoint for " + region + " is not supported yet");
+            }
+            return Region.TERMINAL_API_ENDPOINTS_MAPPING.getOrDefault(region, TERMINAL_API_ENDPOINT_LIVE);
         }
+
+        // Default to EU endpoint if no environment or region is specified
+        return Region.TERMINAL_API_ENDPOINTS_MAPPING.get(Region.EU);
```
Kwok-he-Chu, line 163: Default to `Client.TERMINAL_API_ENDPOINT_TEST` when no environment is specified

Kwok-he-Chu (approved):
Nicely done!

Kwok-he-Chu (approved):

PR #1412, opened 2024-12-22
Deprecate TerminalLocalAPIUnencrypted.java

ayodejidev:
**Description**

This PR:
- adds a deprecation note to `TerminalLocalAPIUnencrypted` class
- provided a developer guide on how to use the local terminal API without encryption in the README
- Refactors tables in the README


src/main/java/com/adyen/service/TerminalLocalAPIUnencrypted.java
```diff
@@ -21,12 +21,16 @@
 import java.security.cert.X509Certificate;
 
 /**
+ * @deprecated This class is planned for deprecation in future releases.
```
gcatanese, line 24: I think a more clear guidance would be: `This class is deprecated and will be removed in on of the next future releases`

README.md
```diff
@@ -488,7 +488,7 @@ When using Attachments, ensure content is provided as a byte array. It's importa
 If you wish to develop the Local Terminal API integration parallel to your encryption implementation, you can opt for the unencrypted version. Be sure to remove any encryption details from the CA terminal config page.
```
gcatanese, line 488: We could add a note here: `Consider this ONLY for development and testing on locahost`

gcatanese (approved):
Good work, approved but I have added few suggestions

src/main/java/com/adyen/service/TerminalLocalAPIUnencrypted.java
ayodejidev: Thanks. Updated the messaage.

DjoykeAbyah (approved):

Issue #1445, opened 2025-03-17
[Feature]: Surcharge is not able to be specified in Checkout API payment requests


sdkindy:
### Problem statement

The appropriate changes have been made to facilitate surcharge with terminals (card-present), but not for card-not-present.

### Proposed solution

Add surcharge to PaymentRequest.

gcatanese:
Thank you for reporting this, it is fixed by #1477

PR #1439, opened 2025-03-04
Code generation: update services and models

AdyenAutomationBot:
OpenAPI spec files or templates have been modified on 01-05-2025   by [commit](https://github.com/Adyen/adyen-openapi/commit/3299c6bd5df2177dd1a649d60f07d99c6baba737).

PR #1446, opened 2025-03-17
Add surcharge property with single property amount to allow checkout …

sdkindy:
…API requests with surcharge

**Description**
There is no ability to specify a surcharge amount for Checkout API payment requests. This adds a `Surcharge` class which contains a single property of `amount` and adds a `Surcharge` instance to `PaymentRequest`.

**Tested scenarios**
Please feel free to test on your end as you have the ability to do so. This was given to our team at Archy as the only option per the Adyen implementation team.

**Fixed issue**: #1445


Kwok-he-Chu:
Hi @sdkindy,

Thank you for the contribution. Unfortunately, we cannot merge this PR as these models are generated from the [Open Api specifications](https://github.com/Adyen/adyen-openapi). To test this feature, we'd recommend compiling the library (locally) for now and use it in your pilot. 

**Next steps on our end:**
* [ ] Once the feature is generally available, we'll include this property in the OpenApi specifications. This will then [trigger our automation](https://github.com/Adyen/adyen-java-api-library/pull/1439) that will include the property.

gcatanese:
Thank you @sdkindy for your identifying the problem and submitting your contribution!!
As explained above, we generate the Java models from the OpenAPI specification, so we are not able to merge your PR (although your PR is correct and it would solve the problem).

`Surcharge` is now included, fixed by #1477

PR #1447, opened 2025-03-20
Ovverride default ApiError message with the error message from the API response

gcatanese:
This PR improves the error handling by refactoring the code that creates the `ApiException`.

In the new approach, the API error (JSON) is parsed first to obtain all the error details, then the `ApiException` object is instantiated using a meaningful error message (instead of a default `HTTP Exception error` message.


ayodejidev (approved):
Looks good. Thanks for making it clearer what the exception represents.

PR #1461, opened 2025-04-29
Initialise List/Set/Map as null

gcatanese:
In the Adyen Java v36 we have upgraded (in Checkout) the OpenAPI Generator: this has introduced a different way of initialising attributes that include container objects (Map, List, Set).

**Before v36**
```  
  private List<PaymentMethodIssuer> issuers = null;
```
**From v36**
```
  private List<PaymentMethodIssuer> issuers = new ArrayList<>();
```

While the above approaches are semantically different, they should be treated as equivalent in terms of information content (no list is the same as an empty list). However the change above has broken the compatibility with the Adyen iOS SDK.

This PR reverts the initialisation of attributes that contain an instance of List/Set/Map.


src/main/java/com/adyen/model/checkout/Donation.java
```diff
@@ -47,8 +48,11 @@ public class Donation {
   public static final String JSON_PROPERTY_MAX_ROUNDUP_AMOUNT = "maxRoundupAmount";
   private Long maxRoundupAmount;
 
+  public static final String JSON_PROPERTY_TYPE = "type";
+  private String type;
```
Kwok-he-Chu, line 52: FYI: donation class now has a type

Kwok-he-Chu (approved):

Kwok-he-Chu (approved):

PR #1471, opened 2025-05-01
Add BalancePlatform WebhookSettings

gcatanese:
PR to add WebhookSettings to support the following endpoints:
* `POST /balancePlatforms/(balancePlatformId)/webhooks/(webhookId)/settings`
* `GET /balancePlatforms/(balancePlatformId)/webhooks/(webhookId)/settings`
* `GET /balancePlatforms/(balancePlatformId)/webhooks/(webhookId)/settings/{settingId}`
* `PATCH /balancePlatforms/(balancePlatformId)/webhooks/(webhookId)/settings/{settingId}`
* `DELETE /balancePlatforms/(balancePlatformId)/webhooks/(webhookId)/settings/{settingId}`


src/main/java/com/adyen/model/balanceplatform/AssociationInitiateRequest.java
```diff
@@ -92,6 +90,9 @@ public AssociationInitiateRequest ids(List<String> ids) {
   }
 
   public AssociationInitiateRequest addIdsItem(String idsItem) {
+    if (this.ids == null) {
```
Kwok-he-Chu, line 93: I don't think we should instantiate the list here, it hides initialization logic (& side effects), if we need it it should be on constructor/initilization level, what do you think?

src/main/java/com/adyen/model/balanceplatform/BankAccountDetails.java
```diff
@@ -46,13 +44,13 @@ public class BankAccountDetails {
   private String accountNumber;
 
   public static final String JSON_PROPERTY_ACCOUNT_TYPE = "accountType";
-  private String accountType;
+  private String accountType = "checking";
```
Kwok-he-Chu, line 47: Is this assignment intended?

src/main/java/com/adyen/model/balanceplatform/AssociationInitiateRequest.java
gcatanese: We have decided to initialise the containers (map/list/set) as `null`, as it used to be in previous versions (see iOS SDK incident). It is also a popular Java practise to initialise the container only if you use it.

I also prefer the initialisation at constructor level to be honest, but given the above we shouldn't change it.

PR #1483, opened 2025-05-14
Avoid null in User-Agent header

gcatanese:
PR that makes sure `User-Agent` header doesn't contain `null` when `applicationName` is undefined

### DIFF
--- a/src/test/java/com/adyen/httpclient/ClientTest.java
+++ b/src/test/java/com/adyen/httpclient/ClientTest.java
@@ -128,8 +128,7 @@ public class ClientTest extends BaseTest {
                 "https://chekout.adyen.com",
                 "{}",
                 new Config()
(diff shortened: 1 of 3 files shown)


src/test/java/com/adyen/httpclient/ClientTest.java
```diff
@@ -41,7 +50,7 @@ public void testConfigLiveClient() {
         config.setLiveEndpointUrlPrefix("prefix");
         config.setApiKey(apiKey);
         Client client = new Client(config);
-        Assert.assertEquals(Environment.LIVE, client.getConfig().getEnvironment());
+        assertEquals(Environment.LIVE, client.getConfig().getEnvironment());
```
Kwok-he-Chu, line 53: boys scout rule: should we also assert that the prefix is correctly set? We can add another test here

PR #1514, opened 2025-07-01
Handling unknown enum values

gcatanese:
When an unknown enum value is found in the JSON payload (API response, webhook request) the deserialization throws an exception. This PR makes the deserialization logic more robust to be able to handle:

- unknown enum values
- unknown Terminal API Notification Events

The Mustache templates have been update to set the attribute as `null`, instead of throwing an `IllegalArgumentException`.

Unit tests have been added to verify the deserialization code.

## Implementation Details

When parsing a JSON payload and encountering unknown enumsm we have 2 options:

1. strict validation throwing `IllegalArgumentException` if the value is unknown
2. don't fail and ignore, hence set the value as null

We have adopted the latter as it allows us to be resilient when new enums appear in the API responses.

We have also decided to keep case-sensitivity when checking enum values, this is how OpenAPI defines enums.


Kwok-he-Chu:
Approved, note that this change does not handle `caseInsensitivity` comparisons when looping over all enum values. Developers will have to handle the `null`-case as well

Kwok-he-Chu (approved):

PR #1522, opened 2025-07-15
Improve generated code in JSON class

gcatanese:
Improve the generated `JSON.java`:

- make variables final (where applicable)
- make constructor private (it is a Singleton)
- replace explicit argument with `<>` 
- set correctly `ALLOW_COERCION_OF_SCALARS`

### DIFF
--- a/templates-v7/libraries/jersey3/JSON.mustache
+++ b/templates-v7/libraries/jersey3/JSON.mustache
@@ -28,10 +28,10 @@ import jakarta.ws.rs.ext.ContextResolver;
 public class JSON implements ContextResolver<ObjectMapper> {
   private static ObjectMapper mapper;
 
-  public JSON() {
+  private JSON() {
     mapper = new ObjectMapper();
     mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    JsonMapper.builder().configure(MapperFeature.ALLOW_COERCION_OF_SCALARS, false);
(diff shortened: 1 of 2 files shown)


templates-v7/libraries/jersey3/JSON.mustache
```diff
@@ -28,10 +28,10 @@ import jakarta.ws.rs.ext.ContextResolver;
 public class JSON implements ContextResolver<ObjectMapper> {
   private static ObjectMapper mapper;
 
-  public JSON() {
+  private JSON() {
     mapper = new ObjectMapper();
     mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    JsonMapper.builder().configure(MapperFeature.ALLOW_COERCION_OF_SCALARS, false);
+    mapper.configure(MapperFeature.ALLOW_COERCION_OF_SCALARS, true);
```
Kwok-he-Chu, line 34: `.configure(MapperFeature.ALLOW_COERCION_OF_SCALARS, false)` This changes the json behavior from a more `strict type of matching` to a more `flexible` matching approach, could you add the reason for this specific change?

gcatanese: The previous line didn't have any effect, as the JsonMapper builder requires invoking the `build()` method once the properties are configured, so basically the default value (`MapperFeature.ALLOW_COERCION_OF_SCALARS = true`) would apply.
The new line doesn't change the behaviour, it only makes clear which value is used for that feature (we could omit it but I think in this way we make it clear).

PR #1542, opened 2025-08-18
Add support for LEM v4

gcatanese:
Adding support for [LEM  v4](https://docs.adyen.com/api-explorer/legalentity/latest/overview).

See our documentation:
- [Onboarding v4](https://docs.adyen.com/platforms/onboard-users/onboarding-versions/onboarding-v4/)
- [API changes v4 vs v3](https://docs.adyen.com/api-explorer/api-diff-tool?api=legalentity&from=3&to=4)

### Breaking Changes 🛠 
- `BusinessLine`: remove `capability`
- `SourceOfFunds`: remove `acquiringBusinessLineId`
- `SourceOfFunds`: make `adyenProcessedFunds` required

### Other Changes 💎 

- `Individual`: add `support`
- `Organization`: several attributes have been added: `support `, `doingBusinessAsAbsent`, `registrationNumberAbsent`, `legalForm`, etc...
- `SoleProprietorship`: add `doingBusinessAsAbsent`, `registrationNumberAbsent`
- `Trust`: add `doingBusinessAsAbsent`
- `UnincorporatedPartnership`: add `doingBusinessAsAbsent`
- `ServiceEnum`: add new value`"issuing"`
- `SourceOfFunds`: add several new attributes (`amount`, `assetMonthsHeld`, etc..)
- `SourceOfFunds`: add array of `Financier`
- `SourceOfFunds. TypeEnum`: add several new values (`"employment"`, `"donations"`, etc..)
- `TaxInformation`: add `numberAbsent`
- `Document`: add `proofOfDirector`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @Adyen/developer-relations
+* @Adyen/developer-relations @Adyen/api-library-maintainers
--- a/src/main/java/com/adyen/Service.java
+++ b/src/main/java/com/adyen/Service.java
@@ -105,8 +105,8 @@ public class Service {
       if (config.getLiveEndpointUrlPrefix() == null) {
         throw new IllegalArgumentException("please provide a live url prefix in the client");
       }
-      if (url.contains("/possdk/v68")) {
-        // Temporary until they fix possdk
+      if (url.contains("/possdk/")) {
+        // Custom handling of POS Mobile API url
         url =
             url.replaceFirst(
                 "https://checkout-test.adyen.com/",
--- a/src/test/java/com/adyen/BaseTest.java
+++ b/src/test/java/com/adyen/BaseTest.java
@@ -32,15 +32,7 @@ import com.adyen.enums.VatCategory;
 import com.adyen.httpclient.AdyenHttpClient;
 import com.adyen.httpclient.HTTPClientException;
 import com.adyen.model.additionalData.InvoiceLine;
-import com.adyen.model.nexo.AmountsReq;
-import com.adyen.model.nexo.MessageCategoryType;
-import com.adyen.model.nexo.MessageClassType;
-import com.adyen.model.nexo.MessageHeader;
-import com.adyen.model.nexo.MessageType;
-import com.adyen.model.nexo.PaymentTransaction;
-import com.adyen.model.nexo.SaleData;
-import com.adyen.model.nexo.SaleToPOIRequest;
-import com.adyen.model.nexo.TransactionIdentification;
+import com.adyen.model.nexo.*;
 import com.adyen.model.payment.ApplicationInfo;
 import com.adyen.model.payment.AuthenticationResultRequest;
 import com.adyen.model.payment.BrowserInfo;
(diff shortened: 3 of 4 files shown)


src/main/java/com/adyen/model/legalentitymanagement/Financier.java
```diff
@@ -0,0 +1,236 @@
...
+  private Amount amount;
+
+  public static final String JSON_PROPERTY_FIRST_NAME = "firstName";
+  private String firstName;
+
+  public static final String JSON_PROPERTY_LAST_NAME = "lastName";
+  private String lastName;
+
+  public static final String JSON_PROPERTY_LOCATION = "location";
+  private String location;
```
galesky-a, line 38: The naming and docs for this property could be improved

src/main/java/com/adyen/model/legalentitymanagement/Individual.java
```diff
@@ -292,6 +296,39 @@ public void setResidentialAddress(Address residentialAddress) {
...
+  }
+
+  /**
+   * Get support
+   *
+   * @return support
+   */
+  @JsonProperty(JSON_PROPERTY_SUPPORT)
+  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
+  public Support getSupport() {
```
galesky-a, line 317: Seems like the description of this method is not matching the Spec.

src/main/java/com/adyen/model/legalentitymanagement/SourceOfFunds.java
```diff
@@ -17,31 +17,99 @@
...
+
+    INSURANCESETTLEMENT(String.valueOf("insuranceSettlement")),
+
+    CRYPTOCURRENCYINCOME(String.valueOf("cryptocurrencyIncome")),
+
+    ASSETSALE(String.valueOf("assetSale")),
+
+    LOANS(String.valueOf("loans")),
+
+    GAMBLINGWINNINGS(String.valueOf("gamblingWinnings"));
```
galesky-a, line 112: Likely breaking to change now but `GAMBLING_WINNINGS` would be better than `GAMBLINGWINNINGS`

src/main/java/com/adyen/model/legalentitymanagement/SourceOfFunds.java
```diff
@@ -239,16 +666,64 @@ public TypeEnum getType() {
...
+   *     **cryptocurrencyIncome** * **assetSale** * **loans** * **gamblingWinnings**
    */
   @JsonProperty(JSON_PROPERTY_TYPE)
   @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
   public void setType(TypeEnum type) {
     this.type = type;
   }
 
+  /**
+   * The location of the gambling site. Required if &#x60;type&#x60; is **gamblingWinnings**. If the
```
galesky-a, line 686: Adding as a note. Let's reword this better to start with the "Required..." part. Since this field is called `website` it's likely to be used for other kinds of websites.

src/main/java/com/adyen/model/legalentitymanagement/Support.java
```diff
@@ -0,0 +1,154 @@
...
+   * @return email The email address of the legal entity.
+   */
+  @JsonProperty(JSON_PROPERTY_EMAIL)
+  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
+  public String getEmail() {
+    return email;
+  }
+
+  /**
+   * The email address of the legal entity.
```
galesky-a, line 54: Adding as a note. This description and others on this file should be reworded to match the scope.

src/main/java/com/adyen/model/legalentitymanagement/Individual.java
gcatanese: The issue seems to be that the OpenAPI Generator doesn't pickup the `description` defined as sibling of the `$ref`
```
        support:
          x-addedInVersion: '4'
          description: Support information for the legal entity.
          $ref: '#/components/schemas/Support'
```
It is only able to get the description within the `Support` component, which unfortunately is not provided.

In OAS 3.0 `$ref` siblings would be ignored, this has changed in OAS 3.1, it is possible that the OpenAPI Generator is still catching up. We will need to investigate if/how we can fix it ourselves.

src/main/java/com/adyen/model/legalentitymanagement/SourceOfFunds.java
gcatanese: I know, but I have checked and we have always generated enums without breaking the words.
If an enum is for example `upi_collect` becomes `UPI_COLLECT`, but `upiCollect` would be `UPI_COLLECT`.

Updating the convention would affect many enums, maybe something to consider in the future?

src/main/java/com/adyen/model/legalentitymanagement/SourceOfFunds.java
gcatanese: This needs to be done in the OpenAPI file

src/main/java/com/adyen/model/legalentitymanagement/Individual.java
galesky-a: Clear, let's make sure to create a ticket to follow up on it

galesky-a (approved):

PR #1603, opened 2025-12-04
Handle nullable properties

### LINKED ISSUE #1584: [Bug]: Nulling out or resetting surcharge configurations for terminal settings does not work
### Description

Adyen terminal settings API does not allow resetting surcharge configurations (so they no longer apply)

### Steps to reproduce

```java
TerminalSettings terminalSettings = new TerminalSettings();
// Toggle off surcharging
terminalSettings.setSurcharge(null);
```

This should work to basically reset surcharge configurations so none apply on the terminal devices assigned to the store. We get a 200 back, but it no-ops and doesn't actually undo the surcharge configurations. 

The suggestion from support was to pass an empty string. Given your library, this would not work or be viable given that the expected value is a `Surcharge` object. 

This is a significant issue.

### Actual behavior

You cannot toggle off surcharging after setting it in terminal settings for a store.

### Expected behavior

You should be able to set and toggle off surcharging; it shouldn't be a one and done user action.

### Code snippet or screenshots (if applicable)

Relevant code block (notice how our expectation as a consumer of this library is that we should be able to toggle on/off surcharging interchangeably):
```
  TerminalSettings terminalSettings = new TerminalSettings();

        FeeStructure feeStructure = practiceBillingSettingDao.getOrCreatePracticeBillingSetting(requestContext)
                .getFeeStructure();

        if (FeeStructure.SURCHARGE.equals(feeStructure)) {
            // Toggle on surcharging
            List<ModelConfiguration> configurations = AdyenConstants.SUPPORTED_PAYMENT_METHODS.stream()
                    .map(paymentMethodSetupInfo -> new ModelConfiguration().brand(paymentMethodSetupInfo.getValue())
                            .currencies(List.of(new Currency().currencyCode("USD")
                                    .percentage(paymentConfig.getPassTheFeePercentageSurcharge().doubleValue())))
                            .sources(List.of(AdyenConstants.ADYEN_CP_TERMINAL_SETTINGS_SURCHARGE_FUNDING_SOURCE)))
                    .toList();

            Surcharge surcharge = new Surcharge();
            surcharge.setConfigurations(configurations);
            surcharge.setAskConfirmation(true);
            terminalSettings.setSurcharge(surcharge);
        } else {
            // Toggle off surcharging
            terminalSettings.setSurcharge(null);
        }

        adyenGateway.updateTerminalSettings(adyenStoreIdentity.getStoreId(), terminalSettings).join();
```

`AdyenGateway#updateTerminalSettings`:

```java
public CompletableFuture<TerminalSettings> updateTerminalSettings(
            @NotNull String storeId,
            @NotNull TerminalSettings terminalSettings) {
        TerminalSettingsStoreLevelApi terminalSettingsStoreLevelApi = new TerminalSettingsStoreLevelApi(
                this.adyenMerchantClient);
        return CompletableFuture.supplyAsync(() -> {
            try {
                return terminalSettingsStoreLevelApi.updateTerminalSettingsByStoreId(storeId, terminalSettings);
            } catch (Exception e) {
                printError(e);
                CompletableFuture<TerminalSettings> failedFuture = new CompletableFuture<>();
                failedFuture.completeExceptionally(e);
                return failedFuture.join(); // this will throw the exception in the calling thread
            }
        });
    }
```

### Adyen Java API Library version

38.0.0.0

### Java version

17.0.12

### Operating System

macOS

### Additional context

This is not an OS-specific issue.

gcatanese:
This pull request adds support for explicit null serialization in the **Management API classes**, allowing clients to control whether attributes with null values are included in the JSON payload. 

The default behavior doesn't change: `null` attributes are not included in the JSON requests. However, developers can override the logic by customizing the request object:

- set `includeNullValues` as `true`
- set explicitly the attribute(s) as `null`  

```
TerminalSettings terminalSettings = new TerminalSettings();
    terminalSettings
        .includeNullValues(true)
        .connectivity(new Connectivity().simcardStatus(Connectivity.SimcardStatusEnum.ACTIVATED))
        .setSurcharge(null);
```
The changes introduce new flags and helper methods to track when attributes are explicitly set to null. This is done by introducing a state (boolean flag) that it is marked as `true` when the attribute is modified. 
Attributes initialised as null are not considered updated and it will not included in the serialization.

This improves API flexibility and consistency, especially for partial updates or PATCH requests.

These changes ensure that API consumers have fine-grained control over the inclusion of null values in their requests, which is important for partial updates (i.e. update `TerminalSettings`) and for conforming to various API contract requirements.

### Other changes

The Management API models have been regenerated to implement the new Mustache templates, meanwhile the Management API OpenAPI spec has changed, therefore there are other changes:

- `AffirmInfo` add `pricePlan` attribute
- `PaymentMethod` add support for `Givex`, `Scs`, `Valuelink`
- `PaymentMethodResponse` add support for `Klarna_b2b`, `Givex`, `Scs`, `Valuelink`
- `PaymentMethodSetupInfo` add support for `Klarna_b2b`, `Givex`, `Scs`, `Valuelink`
- `TerminalSettings` add support for `Moto`
- `UpdatePaymentMethodInfo` deprecate `storeIds` attribute

Fix #1584

### DIFF
--- a/src/test/java/com/adyen/serializer/ModelTest.java
+++ b/src/test/java/com/adyen/serializer/ModelTest.java
@@ -242,4 +242,21 @@ public class ModelTest {
     String json = terminalSettings.toJson();
     assertEquals("{\"connectivity\":{\"simcardStatus\":\"ACTIVATED\"},\"surcharge\":null}", json);
   }
+
+	// test values, that are not explicitly SET as null, are not serialized (even when includeNullValues is true)
+	@Test
+	public void testToJsonFlagTrueButFieldNotSet() throws JsonProcessingException {
+		TerminalSettings terminalSettings = new TerminalSettings();
+		// must consider null values that are explicitly set
+		terminalSettings.includeNullValues(true);
+
+		// We set connectivity, but we NEVER touch 'surcharge'
+		terminalSettings.setConnectivity(new Connectivity().simcardStatus(Connectivity.SimcardStatusEnum.ACTIVATED));
+
+		String json = terminalSettings.toJson();
+
+		// Expectation: 'surcharge' should NOT be present, even though the flag is true.
+		// It should only be present if we called setSurcharge().
+		assertEquals("{\"connectivity\":{\"simcardStatus\":\"ACTIVATED\"}}", json);
+	}
 }


src/main/java/com/adyen/model/balanceplatform/DeviceInfo.java
```diff
@@ -11,145 +11,53 @@
...
-  DeviceInfo.JSON_PROPERTY_ISO_DEVICE_TYPE,
-  DeviceInfo.JSON_PROPERTY_MSISDN,
-  DeviceInfo.JSON_PROPERTY_OS_NAME,
-  DeviceInfo.JSON_PROPERTY_OS_VERSION,
-  DeviceInfo.JSON_PROPERTY_PAYMENT_TYPES,
-  DeviceInfo.JSON_PROPERTY_SERIAL_NUMBER,
-  DeviceInfo.JSON_PROPERTY_STORAGE_TECHNOLOGY
-})
+@JsonPropertyOrder({DeviceInfo.JSON_PROPERTY_FORM_FACTOR, DeviceInfo.JSON_PROPERTY_OS_NAME})
 public class DeviceInfo {
```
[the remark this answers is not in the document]
gcatanese: This change is reverted

src/main/java/com/adyen/model/acswebhooks/Amount.java
```diff
@@ -145,6 +189,33 @@ private String toIndentedString(Object o) {
...
+
+    return nulls;
+  }
+
+  // add to map when value is null
+  private void addIfNull(Map<String, Object> map, String key, Object value) {
+    if (value == null) {
+      map.put(key, null);
+    }
+  }
```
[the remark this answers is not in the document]
gcatanese: Fair point but it requires a significant refactoring (introducing a base class, or helper method in a different class).

src/main/java/com/adyen/model/acswebhooks/Amount.java
```diff
@@ -23,30 +25,47 @@ public class Amount {
...
   private String currency;
 
+  /** Mark when the attribute has been explicitly set. */
+  private boolean isSetCurrency = false;
+
   public static final String JSON_PROPERTY_VALUE = "value";
   private Long value;
 
+  /** Mark when the attribute has been explicitly set. */
+  private boolean isSetValue = false;
```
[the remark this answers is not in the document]
gcatanese: Another valid approach, less duplication but code readability is also affected IMO, being minor I will not consider it

src/test/java/com/adyen/serializer/ModelTest.java
```diff
@@ -195,4 +197,35 @@ public void testFromJsonCheckoutPaymentMethodBcmcMobile() throws Exception {
         StoredPaymentMethodDetails.TypeEnum.BCMC_MOBILE, storedPaymentMethodDetails.getType());
     assertEquals("7219687191761347", storedPaymentMethodDetails.getStoredPaymentMethodId());
   }
+
+  @Test
+  public void testToJsonTerminalSettingsSurchargeNotSet() throws JsonProcessingException {
```
galesky-a, line 203: We are missing another test that the default behavior still works correctly when we explicitly pass `null`:
```java
@Test
public void testToJsonTerminalSettingsSurchargeSetToNullButFlagFalse() throws JsonProcessingException {
    TerminalSettings terminalSettings = new TerminalSettings();
    // We explicitly set null, BUT we keep the flag false (default)
    terminalSettings.includeNullValues(false); 
    terminalSettings.setSurcharge(null);

    String json = terminalSettings.toJson();
    
    // Expectation: The null should be IGNORED because the flag is off.
    // Current tests do not verify this; they only verify "Not Set".
    assertEquals("{\"connectivity\":null}", json); // Or whatever the base empty JSON looks like
}
```

gcatanese: you mean setting `null` without enabling `includeNullValues`, yes, I will add a test

src/test/java/com/adyen/serializer/ModelTest.java
```diff
@@ -195,4 +197,35 @@ public void testFromJsonCheckoutPaymentMethodBcmcMobile() throws Exception {
...
+    TerminalSettings terminalSettings = new TerminalSettings();
+    terminalSettings.setConnectivity(
+        new Connectivity().simcardStatus(Connectivity.SimcardStatusEnum.ACTIVATED));
+    String json = terminalSettings.toJson();
+    assertEquals("{\"connectivity\":{\"simcardStatus\":\"ACTIVATED\"}}", json);
+  }
+
+  @Test
+  public void testToJsonTerminalSettingsSurchargeSetToNull() throws JsonProcessingException {
```
galesky-a, line 224: In both this and the next test, we test that "sucharge" is null when explicitly set **and** the flag is true. But we should also test what happens when NOT explicitly set:

```java
@Test
public void testToJsonFlagTrueButFieldNotTouched() throws JsonProcessingException {
    TerminalSettings terminalSettings = new TerminalSettings();
    terminalSettings.includeNullValues(true);
    
    // We set connectivity, but we NEVER touch 'surcharge'
    terminalSettings.setConnectivity(new Connectivity().simcardStatus(Connectivity.SimcardStatusEnum.ACTIVATED));

    String json = terminalSettings.toJson();

    // Expectation: 'surcharge' should NOT be present, even though the flag is true.
    // It should only be present if we called setSurcharge().
    assertEquals("{\"connectivity\":{\"simcardStatus\":\"ACTIVATED\"}}", json);
}
```

src/test/java/com/adyen/serializer/ModelTest.java
gcatanese: Done

src/test/java/com/adyen/serializer/ModelTest.java
gcatanese: Good point, another scenario to cover. Done

galesky-a (approved):

PR #1605, opened 2025-12-05
Generate Transfers API and TransferWebhooks models

gcatanese:
PR to generate the Transfers API and TransferWebhooks models.

### Transfers API
- In `UltimatePartyIdentification` add `fundingInstrument` attribute
- In `BankAccountV3` add `storedPaymentMethodId` attribute
- In `Transfer`, `TransferData` and `TransferEvent` add add enum `ReasonEnum.SCHEMEADVICE`

### Transfer Webhooks
- In `BankAccountV3` add `storedPaymentMethodId` attribute
- In `TransferEvent` and `TransferData` add add enum `ReasonEnum.SCHEMEADVICE`

All models have been updated to handle the nullable value (Mustache templates have been updated as part of https://github.com/Adyen/adyen-java-api-library/pull/1603)

### DIFF
--- a/src/main/java/com/adyen/model/transferwebhooks/Address.java
+++ b/src/main/java/com/adyen/model/transferwebhooks/Address.java
@@ -11,8 +11,6 @@
 
 package com.adyen.model.transferwebhooks;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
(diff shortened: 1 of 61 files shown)


src/main/java/com/adyen/model/transfers/Transaction.java
```diff
@@ -668,6 +761,66 @@ private String toIndentedString(Object o) {
...
 
+  /** Returns a map of properties to be merged into the JSON payload as explicit null values. */
+  @JsonInclude(JsonInclude.Include.ALWAYS)
+  @JsonAnyGetter
+  public Map<String, Object> getExplicitNulls() {
+    if (!this.includeNullValues) {
+      return Collections.emptyMap();
+    }
+
+    Map<String, Object> nulls = new HashMap<>();
```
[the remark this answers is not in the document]
gcatanese: Uhm, maybe in a future iteration

src/main/java/com/adyen/model/transferwebhooks/Address.java
```diff
@@ -30,21 +32,45 @@ public class Address {
...
   private String stateOrProvince;
 
+  /** Mark when the attribute has been explicitly set. */
+  private boolean isSetStateOrProvince = false;
+
+  /**
+   * Sets whether attributes with null values should be explicitly included in the JSON payload.
+   * Default is false.
+   */
+  @JsonIgnore private boolean includeNullValues = false;
```
galesky-a: Would it be easy to not add this to webhooks actually ? They don't make a lot of sense since the SDK would be in the receiving end

gcatanese: True. The advantage now is consistent generation across all models, but we can customise it to target only the API models. It introduces a custom behavior, but this code is typically easy to understand and we can add tests to check the generated code.

PR #1644, opened 2026-01-05
[balanceplatform] Code generation: update services and models

AdyenAutomationBot:
This PR contains the automated changes for the `balanceplatform` service.

The commit history of this PR reflects the `adyen-openapi` commits that have been applied.


src/main/java/com/adyen/model/balanceplatform/Balance.java
jeandersonbc, line 1: Pending Available removed (because...?) and not used else where at service level.

src/main/java/com/adyen/model/balanceplatform/DeviceInfo.java
jeandersonbc, line 1: phone info added

PR #1660, opened 2026-01-09
feat(ci): create in place formatting workflow for composable steps

galesky-a:
**Description**
  This PR refactors the formatting workflow to use a composite action. This allows other workflows to format the code in-place before calculating a diff, preventing spurious diffs from formatting changes.

This is part of a larger effort to fix the generation and formatting flow across all langs as described [here](https://github.com/Adyen/adyen-sdk-automation/pull/85)

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,12 +1,7 @@
-# Deprecated
-
 name: Format and Commit Java Code
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - 'sdk-automation/*'
 
 permissions:
   contents: write


.github/workflows/format.yml
```diff
@@ -1,4 +1,6 @@
-name: Java Code Format
+# Deprecated
+
+name: Format and Commit Java Code
```
gcatanese, line 1: The workflow should no longer run `on push`, right? The code has been already formatted. It is useful to leave the `workflow_dispatch` to run it on demand

.github/workflows/format.yml
```diff
@@ -1,4 +1,6 @@
-name: Java Code Format
+# Deprecated
```
gcatanese: I don't think it is deprecated, it can be used on demand (in probably specific cases - say we want to format test classes)

PR #1661, opened 2026-01-12
fix(ci): ensure release workflow triggers only on merged requests

galesky-a:
This pull request fixes an issue in the release workflow that could cause it to trigger on any closed or merged pull request. The workflow has been updated to ensure that releases are only triggered when a pull request is merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,8 @@
-name: Release
+name: Automated Release Preparation
 
+# This workflow runs automatically when a pull request is merged into the main branch.
+# It prepares the next release by creating or updating a "Release PR" with the new version and changelog.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,5 +1,8 @@
-name: Release
+name: Automated Release Preparation
 
+# This workflow runs automatically when a pull request is merged into the main branch.
```
thomasc-adyen, line 3: Hey @galesky-a, is my understanding correct that there will always be only one open release PR for a given version?

galesky-a: As long as there is at least 1 commit between the last Tag (release) and `main`, yes. There is always be a PR like [this](https://github.com/Adyen/adyen-java-api-library/pull/1579):

PR #1732, opened 2026-02-23
test(terminal): validate if boxed flags will be passed as nullable

galesky-a:
**Description**

**Tested scenarios**

**Fixed issue**:


jeandersonbc:
@galesky-a  is this still relevant? Otherwise, I think we should close it.

PR #1802, opened 2026-03-25
Remove javaxb dependency from HMACValidator

jeandersonbc:
**Description**
Instead of using javaxb dependency, we can rely on the existing commons-codec for Hex operations.
This is needed for the removal of the javaxb dependency.
The changes are just a patch to the existing code.

In addition, the pom.xml was cleaned:
- removal of unused plugin
- reorganized dependencies and versions

**Tested scenarios**
- existing tests must pass

### DIFF
--- a/src/main/java/com/adyen/util/HMACValidator.java
+++ b/src/main/java/com/adyen/util/HMACValidator.java
@@ -14,7 +14,7 @@
  *
  * Adyen Java API Library
  *
- * Copyright (c) 2017 Adyen B.V.
+ * Copyright (c) 2026 Adyen B.V.
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
  */
@@ -31,8 +31,9 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
-import javax.xml.bind.DatatypeConverter;
+import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.codec.binary.Hex;
 
 /** Utility class for generating and validating HMAC signatures used in Adyen webhooks. */
 public class HMACValidator {
@@ -48,7 +49,7 @@ public class HMACValidator {
    * @param data the data to sign
    * @param key the HMAC key in hexadecimal format
    * @return the Base64-encoded HMAC signature
-   * @throws IllegalArgumentException if the data or key is null
+   * @throws IllegalArgumentException if the data or key is null or if key is invalid
    * @throws SignatureException if signature generation fails
    */
   public String calculateHMAC(String data, String key)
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -23,16 +23,45 @@
...
-
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <swagger-core-version>1.6.16</swagger-core-version>
         <sonar.organization>adyen</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+
+        <!-- Dependency Versions -->
+        <!-- Compile -->
+        <com.fasterxml.jackson-version>2.21.0</com.fasterxml.jackson-version>
```
thomasc-adyen, line 36: just noticing this, but why do we have `gson` and `jackson` in the same project?

gcatanese: GSON only used by the TerminalAPI

jeandersonbc: Yeah, I'm also curious about this. We do use it for serialization and deserialization in a few places (`.toJson`/`.fromJson`):

- See [TerminalCloudAPI](https://github.com/Adyen/adyen-java-api-library/blob/main/src/main/java/com/adyen/service/TerminalCloudAPI.java#L71-L83)
- There is a [TerminalAPIGsonBuilder](https://github.com/Adyen/adyen-java-api-library/blob/main/src/main/java/com/adyen/terminal/serialization/TerminalAPIGsonBuilder.java) that  relies on javax.xml.bind to do this configuration.

I think as part of the bigger migration, we want to stick with a single JSON library.

src/main/java/com/adyen/util/HMACValidator.java
```diff
@@ -61,7 +61,7 @@ public String calculateHMAC(String data, String key)
         throw new IllegalArgumentException("HMAC key is not provided");
       }
 
-      byte[] rawKey = DatatypeConverter.parseHexBinary(key);
+      byte[] rawKey = Hex.decodeHex(key);
```
thomasc-adyen, line 65: `Hex.decodeHex(key)` throws `DecoderException`, which is currently caught by the broad catch (Exception e) and rethrown as `SignatureException`. This changes public behavior from input-validation failure (`IllegalArgumentException`) to cryptographic failure (`SignatureException`) for invalid hex keys, which is a backward-incompatible regression for callers that handle these cases differently.

Finding from 🤖 Droid - I think it is worth catching it in the same block as the `IllegalArgumentException` so we don't change behavior for callers

pom.xml
thomasc-adyen: Answered my own question 😅 

 This library uses both serialization stacks for historical and compatibility reasons:
   •  Jackson is used broadly by the generated API models (com.adyen.model.*.JSON classes with ObjectMapper).
   •  Gson is still used in specific flows, especially Terminal API and webhook notification helpers (for example TerminalCloudAPI and NotificationRequest#fromJson/toJson).

_Edit: forgot to refresh 😆_

src/main/java/com/adyen/util/HMACValidator.java
jeandersonbc, line 79: exceptions already being addressed here (@thomasc-adyen @gcatanese)

src/main/java/com/adyen/util/HMACValidator.java
```diff
@@ -71,6 +72,8 @@ public String calculateHMAC(String data, String key)
       return new String(Base64.encodeBase64(rawHmac));
     } catch (IllegalArgumentException e) {
       throw new IllegalArgumentException("Missing data or key: " + e.getMessage());
+    } catch (DecoderException e) {
+      throw new IllegalArgumentException("Invalid Hex HMAC key: " + key);
```
jeandersonbc, line 76: Added suggestion from @gcatanese

PR #1806, opened 2026-03-26
Add test coverage for serialization/deserialization for upcoming refactoring

jeandersonbc:
**Description**

In the scope of JAXB migration, we need some sanity check to verify that the changes will work.
This is not a full complete set, but I sampled a few classes to add coverage.

It would be helpful to get some feedback about the tests and whether we could improve something for the upcoming work.

**Tested scenarios**
- simple serialization and deserialization of some impacted classes


gcatanese (approved):
I think the Gemini comments are not critical, we can go ahead

PR #1807, opened 2026-03-27
Removal of JAXB annotations

jeandersonbc:
**Description**

Removal the JAXB-bind related annotations. This change is needed due to Jakarta EE +9 compatibility.
In addition, we Improved type safety for `AuthenticatedData.keyTransportOrKEK` deserialization**:

Previously, items in the keyTransportOrKEK list were deserialized as raw Object (a Gson LinkedTreeMap at runtime), requiring brittle map access to read fields. The `TerminalAPIGsonBuilder` now registers a custom adapter that detects the concrete type of each list item at deserialization time — objects containing a KEKIdentifier field are deserialized as KEK, all others as KeyTransport. This makes safe casting possible:

```java
 // After
 AuthenticatedData data = gson.fromJson(json, AuthenticatedData.class);

 // Cast to KEK
 KEK kek = (KEK) data.getKeyTransportOrKEK().get(0);
 String keyId = kek.getKEKIdentifier().getKeyIdentifier();

 // Cast to KeyTransport
 KeyTransport kt = (KeyTransport) data.getKeyTransportOrKEK().get(0);
 RecipientIdentifier recipient = kt.getRecipientIdentifier();
```

## Further details:

Since the underlying serializer relies on Gson lib, we are unable to use directly Jackson annotations since they don't work as expected. This was verified via test coverage. Furthermore, the `required=true` validation was never enforced (see AbortRequestTest tests).

The existing code in `com.adyen.terminal.serialization` relies on GSON; thus,the JAXB annotations were replaced by the GSON's @SerializedName.

The migration was guided by test coverage and focused on serialization and deserialization of JSON payloads.

POM XML changes:
 - removal of jaxb dependency
 - addition of plugin enforcement to block unwanted dependencies

**Tested scenarios**
- Serialization/Deserialization from external file
- Missing required fields (that are not enforced in practice)


src/main/java/com/adyen/terminal/serialization/TerminalAPIGsonBuilder.java
```diff
@@ -35,20 +33,7 @@ private TerminalAPIGsonBuilder() {}
...
-          } else if (field.getAnnotation(XmlAttribute.class) != null) {
-            XmlAttribute xmlAttribute = field.getAnnotation(XmlAttribute.class);
-            return xmlAttribute.name();
-          } else if (field.getAnnotation(XmlRootElement.class) != null) {
-            XmlRootElement xmlRootElement = field.getAnnotation(XmlRootElement.class);
-            return xmlRootElement.name();
-          }
-          return field.getName();
-        };
+    FieldNamingStrategy fieldNamingStrategy = Field::getName;
```
gcatanese: This can be removed


> `FieldNamingStrategy` is now a no-op (`TerminalAPIGsonBuilder.java`)
> The strategy is set to Field::getName, which is Gson's default. Since all fields now use @SerializedName (which takes precedence), the strategy is dead code. The risk is that any future field added 
>    without @SerializedName will silently fall back to the Java field name instead of the expected PascalCase JSON key. Recommend removing the setFieldNamingStrategy call entirely to make the contract explicit.
>

src/main/java/com/adyen/model/nexo/AuthenticatedData.java
```diff
@@ -37,33 +33,26 @@
...
-@XmlType(
-    name = "AuthenticatedData",
-    propOrder = {"keyTransportOrKEK", "macAlgorithm", "encapsulatedContent"})
 public class AuthenticatedData {
 
   /** The Key transport or kek. */
-  @XmlElements({
-    @XmlElement(name = "KeyTransport", type = KeyTransport.class),
-    @XmlElement(name = "KEK", type = KEK.class)
-  })
```
gcatanese, line 50: Just a remark: `keyTransportOrKEK` starts with a lower case, unlike other attributes.

It is worth checking that we don't move from a case-insensitive approach (with ` XmlType`) to case-sensitive (with `SerializedName`)

src/main/java/com/adyen/model/nexo/AuthenticatedData.java
```diff
@@ -37,33 +33,26 @@
...
+  @SerializedName("keyTransportOrKEK")
   protected List<Object> keyTransportOrKEK;
```
gcatanese, line 40: It looks like here we can improve the deserialization, even if we deal with a polymorphic list

> `AuthenticatedData.keyTransportOrKEK` loses type discrimination (`AuthenticatedData.java`)
>   The old @XmlElements annotation handled KeyTransport.class vs KEK.class polymorphic deserialization. With Gson, items in this list are deserialized as LinkedTreeMap (generic objects), not concrete types. 
   The test explicitly acknowledges this regression with assertInstanceOf(Object.class, ...). Any caller doing (KeyTransport) keyTransportOrKEK.get(0) will get a ClassCastException at runtime. Fix: register 
   a custom TypeAdapter or JsonDeserializer that inspects JSON keys (e.g., presence of KEKIdentifier vs RecipientIdentifier) to determine the concrete type.

It look like before this used to work
```
     AuthenticatedData data = ...; // from JAXB unmarshalling
     KeyTransport kt = (KeyTransport) data.getKeyTransportOrKEK().get(0); // safe cast
```
We can maybe create a test to confirm it.

The fix is a custom `JsonDeserializer` for `AuthenticatedData` (i.e. creating `AuthenticatedDataDeserializer`, registered in `TerminalAPIGsonBuilder`.

jeandersonbc: Perhaps we can try it out on [this](https://github.com/Adyen/adyen-java-api-library/pull/1807/changes#diff-fa5581b71708108056050f2dcae7d16a57e0f7d18414f5972055eef0be697d2cR18-R44) unit test... I'm not so sure this cast worked before but we can
(1) double-check this from master
(2) use [this](https://github.com/Adyen/adyen-java-api-library/pull/1807/changes#diff-fa5581b71708108056050f2dcae7d16a57e0f7d18414f5972055eef0be697d2cR18-R44) test case to update the behavior

I'm gonna take a look a it!

src/main/java/com/adyen/model/nexo/AuthenticatedData.java
jeandersonbc: Good catch, we can try this out with [this payload](src/test/resources/mocks/terminal-api/authenticated-data-kek.json) and some test using it. The expectation is that everything different from` keyTransportOrKEK` should be ignored as an unknown field.

jeandersonbc: I'm catching-up back on this. The current behavior is that fields must match exactly how defined in `@SerializedName`. There was no validation being used for what JAXB annotations used because the underlying Gson lib completely ignores it - it was never verified.

To confirm this, I had reverted changes in `src/main` with latest master and tests remain passing.

src/main/java/com/adyen/model/nexo/AuthenticatedData.java
jeandersonbc: Given that JAXB was ignored by Gson, I'm unsure whether this cast was possible before. In any case, client code should look a lot cleaner with this suggestion: see https://github.com/Adyen/adyen-java-api-library/pull/1807/changes#diff-fa5581b71708108056050f2dcae7d16a57e0f7d18414f5972055eef0be697d2cR32-R35

I think this could be highlighted for future release notes:

**Improved type safety for `AuthenticatedData.keyTransportOrKEK` deserialization**

Previously, items in the keyTransportOrKEK list were deserialized as raw Object (a Gson LinkedTreeMap at runtime), requiring brittle map access to read fields. The `TerminalAPIGsonBuilder` now registers a custom adapter that detects the concrete type of each list item at deserialization time — objects containing a KEKIdentifier field are deserialized as KEK, all others as KeyTransport. This makes safe casting possible:

```java
 // After
 AuthenticatedData data = gson.fromJson(json, AuthenticatedData.class);

 // Cast to KEK
 KEK kek = (KEK) data.getKeyTransportOrKEK().get(0);
 String keyId = kek.getKEKIdentifier().getKeyIdentifier();

 // Cast to KeyTransport
 KeyTransport kt = (KeyTransport) data.getKeyTransportOrKEK().get(0);
 RecipientIdentifier recipient = kt.getRecipientIdentifier();
```

gcatanese: Nice, I like the tests to confirm it 👍

PR #1835, opened 2026-03-31
Handle Cloud Device API BaseURL logic

gcatanese:
## Summary

Extends the `createBaseURL` logic in `Service.java` to properly handle the **Cloud Device API** base URL when switching from test to live environments.

Previously, the generic `-test` to `-live` replacement was applied to all URLs. The Device API uses a different URL pattern (`device-api-test.adyen.com`) that requires region-aware routing in live environments.

## Changes

- **`Service.java`**: Added a condition to detect Device API URLs (`device-api-`) and replace the test URL with the correct live URL based on `terminalApiRegion`:
  - No region or EU region → `https://device-api-live.adyen.com`
  - Other regions (AU, US, APSE) → `https://device-api-live-{region}.adyen.com`

- **`ServiceTest.java`**: Added unit tests covering all region cases (default/EU, AU, US, APSE) and the test environment passthrough.

- **`CloudDeviceApiTest.java`**: Added integration-style tests verifying the resolved `baseURL` field on `CloudDeviceApi` for various environment/region combinations.

### DIFF
--- a/src/test/java/com/adyen/ServiceTest.java
+++ b/src/test/java/com/adyen/ServiceTest.java
@@ -4,8 +4,12 @@ import static org.junit.jupiter.api.Assertions.*;
 
 import com.adyen.enums.Environment;
 import com.adyen.enums.Region;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 /** Tests for {@link Service#createBaseURL(String)}. */
 public class ServiceTest extends BaseTest {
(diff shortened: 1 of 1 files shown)


src/main/java/com/adyen/Service.java
```diff
@@ -139,6 +140,20 @@ protected String createBaseURL(String url) {
       }
     }
 
+    if (url.contains("device-api-")) {
+      if (config.getTerminalApiRegion() == null
+          || config.getTerminalApiRegion().equals(Region.EU)) {
```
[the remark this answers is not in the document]
gcatanese: `Region.IN` to be removed in a different PR

src/main/java/com/adyen/Service.java
```diff
@@ -139,6 +140,20 @@ protected String createBaseURL(String url) {
...
+        return url.replaceFirst(
+            "https://device-api-test.adyen.com", "https://device-api-live.adyen.com");
+      } else {
+        return url.replaceFirst(
+            "https://device-api-test.adyen.com",
+            String.format(
+                "https://device-api-live-%s.adyen.com",
+                config.getTerminalApiRegion().name().toLowerCase()));
+      }
+    }
```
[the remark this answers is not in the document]
gcatanese: Fixed first comment, created #1838 for second comment

src/main/java/com/adyen/Service.java
```diff
@@ -139,6 +140,22 @@ protected String createBaseURL(String url) {
...
+          || config.getTerminalApiRegion().equals(Region.EU)) {
+        url =
+            url.replaceFirst(
+                "https://device-api-test.adyen.com", "https://device-api-live.adyen.com");
+      } else {
+        url =
+            url.replaceFirst(
+                "https://device-api-test.adyen.com",
+                String.format(
+                    "https://device-api-live-%s.adyen.com",
```
thomasc-adyen, line 154: @gcatanese so EU endpoints don't have a region suffix but anything non-EU does?

src/test/java/com/adyen/ServiceTest.java
```diff
@@ -122,4 +123,71 @@ public void testLiveRecurringUrlWithoutPrefix() {
...
+  public void testDeviceApiLiveUrlDefaultRegion() {
+    String testUrl = "https://device-api-test.adyen.com/device/v1";
+    String expectedUrl = "https://device-api-live.adyen.com/device/v1";
+
+    String actualUrl = service.createBaseURL(testUrl);
+    assertEquals(expectedUrl, actualUrl);
+  }
+
+  @Test
+  public void testDeviceApiLiveUrlEuRegion() {
```
thomasc-adyen: For the region tests, it might be more maintainable if we use `@ParametrizedTest` with an `@EnumSource`

src/main/java/com/adyen/Service.java
gcatanese: Correct, see[ API Explorer](https://docs.adyen.com/api-explorer/terminal-api/1/overview)

src/test/java/com/adyen/ServiceTest.java
gcatanese: I have used `@MethodSource` as it seems more suitable than `@EnumSource` because there is no enum for `EU` scenario. The method `deviceApiRegionToLiveUrl` covers all cases and makes it very intuitive.

src/test/java/com/adyen/ServiceTest.java
```diff
@@ -122,4 +127,41 @@ public void testLiveRecurringUrlWithoutPrefix() {
...
+  @Test
+  public void testDeviceApiTestEnvironment() {
+    config.setEnvironment(Environment.TEST);
+    String testUrl = "https://device-api-test.adyen.com/device/v1";
+
+    String actualUrl = service.createBaseURL(testUrl);
+    assertEquals(testUrl, actualUrl);
+  }
+
+  static Stream<Arguments> deviceApiRegionToLiveUrl() {
```
thomasc-adyen, line 140: Nice 👏

PR #1909, opened 2026-04-13
Cloud device API doc

gcatanese:
Update README to add new Cloud Device API
Created `doc` folder to separate `Terminal API` and `Cloud device API` doc
Added a section `Helper classes` in the `README` (not the helper doesn't yet exist, it will included in a different PR)
Added Migration Guide to be published with the Release

**Note**: Terminal API is not marked as deprecated, at this stage the approach is to promote the new Cloud device API for new integrations, making sure customers know that the existing integrations are not affected

### DIFF
--- a/doc/CloudDeviceApi.md
+++ b/doc/CloudDeviceApi.md
@@ -1,0 +1,156 @@
+# Cloud Device API
+
+The [Cloud Device API](https://docs.adyen.com/api-explorer/cloud-device-api/1/overview) is our solution to create best-in-class in-person payments integrations.
+
+With the Cloud device API you can:
+
+- send Terminal API requests to a cloud endpoint. You can use this communication method when it is not an option to send Terminal API requests over your local network directly to a payment terminal.
+- check the cloud connection of a payment terminal or of a device used in a Mobile solution for in-person payments.
+
+## Benefits of the Cloud Device API
+
+The Cloud Device API offers the following benefits:
+- access to API logs in the Customer Area for troubleshooting errors
+- using a version strategy for the API endpoints for controlled and safer rollouts
+- improved reliability and security (OAuth support)
+
+New features and products will be released exclusively on the Cloud Device API
+
+## Use the Cloud Device API
+
+### Setup
+
+First you must initialise the Client **setting the closest** [Region](https://docs.adyen.com/point-of-sale/design-your-integration/terminal-api/#cloud):
+``` java
+// Import the required classes
+import com.adyen.Client;
+import com.adyen.enums.Environment;
+import com.adyen.service.checkout.PaymentsApi;
+import com.adyen.model.checkout.*;
+
+// Setup Client and Service
+Client client = new Client("Your X-API-KEY", Environment.TEST);
+CloudDeviceApi cloudDeviceApi = new CloudDeviceApi(client);
+
+```
+
+### Send a payment SYNC request
+
+```java
+
+SaleToPOIRequest saleToPOIRequest = new SaleToPOIRequest();
+
+MessageHeader messageHeader = new MessageHeader();
+    messageHeader.setProtocolVersion("3.0");
+    messageHeader.setMessageClass(MessageClassType.SERVICE);
+    messageHeader.setMessageCategory(MessageCategoryType.PAYMENT);
+    messageHeader.setMessageType(MessageType.REQUEST);
+    messageHeader.setSaleID("001");
+    messageHeader.setServiceID("001");
+    messageHeader.setPOIID("P400Plus-123456789");
+
+    saleToPOIRequest.setMessageHeader(messageHeader);
+
+PaymentRequest paymentRequest = new PaymentRequest();
+
+SaleData saleData = new SaleData();
+TransactionIdentification transactionIdentification = new TransactionIdentification();
+    transactionIdentification.setTransactionID("001");
+OffsetDateTime timestamp = OffsetDateTime.now(ZoneOffset.UTC);
+    transactionIdentification.setTimeStamp(timestamp);
+    saleData.setSaleTransactionID(transactionIdentification);
+
+PaymentTransaction paymentTransaction = new PaymentTransaction();
+AmountsReq amountsReq = new AmountsReq();
+    amountsReq.setCurrency("EUR");
+    amountsReq.setRequestedAmount(BigDecimal.ONE);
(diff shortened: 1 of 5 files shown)


doc/CloudDeviceApi.md
```diff
@@ -0,0 +1,178 @@
...
+- using a version strategy for the API endpoints for controlled and safer rollouts
+- improved reliability and security (OAuth support)
+
+New features and products will be released exclusively on the Cloud device API
+
+## Use the Cloud device API
+
+### Setup
+
+First you must initialise the Client **setting the closest** [Region](https://docs.adyen.com/point-of-sale/design-your-integration/terminal-api/#cloud):
```
thomasc-adyen: Shouldn't we show setting the region in the example?

thomasc-adyen: We should probably have one example for `test` and one example for `live`

doc/CloudDeviceApi.md
```diff
@@ -0,0 +1,178 @@
...
+### Verify the status of the terminals
+
+
+The Cloud device API allows your integration to check the status of the terminals.
+
+```java
+
+// list of payment terminals or SDK mobile installation IDs
+ConnectedDevicesResponse connectedDevices = cloudDeviceApi.getConnectedDevices("myMerchant");
+System.out.println(connectedDevices.getUniqueDeviceIds());
```
thomasc-adyen, line 135: Should we add a comment to show the shape of the response? (what would actually be printed)

doc/CloudDeviceApi.md
```diff
@@ -0,0 +1,178 @@
...
+### Protect cloud communication
+
+The Adyen Java library supports encrypting request and response payloads, allowing you to secure communication between your integration and the cloud.
+
+```java
+
+//  Encryption credentials from the Terminal configuration on CA
+EncryptionCredentialDetails encryptionCredentialDetails =
+    new EncryptionCredentialDetails()
+        .adyenCryptoVersion(0)
```
thomasc-adyen: Do we actually support a version 0 for both of these?

doc/CloudDeviceApi.md
```diff
@@ -0,0 +1,178 @@
...
+        .keyIdentifier("CryptoKeyIdentifier12345")
+        .keyVersion(0)
+        .passphrase("p@ssw0rd123456");
```
thomasc-adyen, line 174: Should we suggest here loading these via an environment variable? some merchants may think this means it's ok to hardcode the passphrase

doc/MigratingToCloudDeviceApi.md
```diff
@@ -0,0 +1,429 @@
...
+
+## Benefits of the Cloud device API
+
+The Cloud device API introduces several improvements over the Terminal (Cloud) API:
+
+- **API logs in the Customer Area**: troubleshoot errors using the API logs available in your Adyen Customer Area.
+- **Versioned endpoints**: the API uses a version strategy for controlled and safer rollouts.
+- **Improved security**: supports OAuth authentication alongside API key authentication.
+- **Device management endpoints**: query connected devices and check their status directly from your integration.
+- **New features**: future In-Person Payments features and products will be released exclusively on the Cloud device API.
```
thomasc-adyen, line 19: Can we move this to the top? I think this will be the biggest draw for merchants

doc/MigratingToCloudDeviceApi.md
```diff
@@ -0,0 +1,429 @@
...
+| `POIData`  | `setPOIReconciliationID()`  | `setPoIReconciliationID()`  |
+| `CardData` | `getMaskedPAN()`            | `getMaskedPan()`            |
+
+### Type changes
+
+Some field types differ in the generated models.
+
+#### Timestamp fields: `XMLGregorianCalendar` -> `OffsetDateTime`
+
+The most common change. The Cloud device API models use `java.time.OffsetDateTime` for timestamp fields, whereas the Terminal (Cloud) API models use `XMLGregorianCalendar`.
```
thomasc-adyen, line 130: Can we add a bit more explaining the distinction between the two and what the impact will be on how the same timestamp is interpreted?

doc/MigratingToCloudDeviceApi.md
```diff
@@ -0,0 +1,429 @@
...
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+
+TransactionIDType transactionIDType = new TransactionIDType();
+transactionIDType.setTimeStamp(OffsetDateTime.now(ZoneOffset.UTC));
+```
+
+#### Date fields: `String` -> `LocalDate`
+
+Some date fields (e.g. `Instalment.firstPaymentDate`) use `java.time.LocalDate` in the Cloud device API instead of `String`.
```
thomasc-adyen, line 154: Can we emphasize that using `LocalDate` means it's important that the systems the SDK is run on has the correct timezone set?

doc/MigratingToCloudDeviceApi.md
```diff
@@ -0,0 +1,429 @@
...
+
+**Cloud device API:**
+```java
+import java.time.LocalDate;
+instalment.setFirstPaymentDate(LocalDate.of(2025, 1, 15));
+```
+
+#### Numeric fields
+
+- `BigInteger` fields are `Integer` in the Cloud device API
```
thomasc-adyen, line 173: Out of curiosity, is there a risk this creates issues? this is a narrowing conversion

doc/CloudDeviceApi.md
gcatanese: Done, also corrected the snippets

doc/MigratingToCloudDeviceApi.md
gcatanese: Yes, there is always a risk "narrowing" the conversion, but practically `Integer` is big enough for the type of data  we handle. It is most likely that former SDK code didn't really need to use `BigInteger`

doc/CloudDeviceApi.md
gcatanese: No, I think it is a mistake, changed now to `1`

doc/CloudDeviceApi.md
gcatanese: Fair point, but all `README` files in the libraries use the same convention. I suggest to create a new issue if we want to include best practices in the snippets across all libraries.

jeandersonbc (approved):

PR #1913, opened 2026-04-14
Add TAPI utility helpers for PredefinedContent and SaleToAcquirerData

gcatanese:
## Summary

Introduces hand-written utility helpers in `com.adyen.util.tapi` to work with fields from the auto-generated TAPI models that require additional parsing logic:

- `PredefinedContent.ReferenceID` string  with URL query string format 
- `SaleData.SaleToAcquirerData` string with either JSON or key-value pairs

### New classes

**`PredefinedContentHelper`**
Parses the `PredefinedContent.ReferenceID` URL query string into typed accessors: `getEvent()`, `getTransactionId()`, `getTimeStamp()`, and `get(key)`.

**`SaleToAcquirerData`**
A Jackson-based POJO representing the decoded acquirer data. Supports both formats described in the [Adyen docs](https://docs.adyen.com/point-of-sale/add-data):
- Base64-encoded JSON (`fromBase64`)
- Form-encoded key-value pairs (`fromKeyValuePairs`)

**`SaleDataHelper`**
Wraps a `tapi.SaleData` instance and exposes `getSaleToAcquirerData()`, auto-detecting the format of the underlying string.

### Notes
- All helpers live outside `com.adyen.model.tapi` which is auto-generated from the OpenAPI spec
- No dependency on the legacy `com.adyen.model.terminal` package
- Uses `java.util.Base64` and Jackson throughout, consistent with the rest of the TAPI stack

### DIFF
--- a/src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
+++ b/src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
@@ -1,0 +1,148 @@
+package com.adyen.util.tapi;
+
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A helper class to parse and manage the key-value pairs within a PredefinedContent referenceID
+ * string. The referenceID is expected to be in a URL query string format (e.g., {@code
+ * key1=value1&key2=value2}).
+ */
+public final class PredefinedContentHelper {
+
+  private static final String KEY_EVENT = "event";
+  private static final String KEY_TRANSACTION_ID = "TransactionID";
+  private static final String KEY_TIME_STAMP = "TimeStamp";
+
+  /**
+   * Defines the supported events for display notifications within a PredefinedContent reference ID.
+   */
+  public enum DisplayNotificationEvent {
+    TENDER_CREATED,
+    CARD_INSERTED,
+    CARD_PRESENTED,
+    CARD_SWIPED,
+    WAIT_FOR_APP_SELECTION,
+    APPLICATION_SELECTED,
+    ASK_SIGNATURE,
+    CHECK_SIGNATURE,
+    SIGNATURE_CHECKED,
+    WAIT_FOR_PIN,
+    PIN_ENTERED,
+    PRINT_RECEIPT,
+    RECEIPT_PRINTED,
+    CARD_REMOVED,
+    TENDER_FINAL,
+    ASK_DCC,
+    DCC_ACCEPTED,
+    DCC_REJECTED,
+    ASK_GRATUITY,
+    GRATUITY_ENTERED,
+    BALANCE_QUERY_STARTED,
+    BALANCE_QUERY_COMPLETED,
+    LOAD_STARTED,
+    LOAD_COMPLETED,
+    PROVIDE_CARD_DETAILS,
+    CARD_DETAILS_PROVIDED
+  }
+
+  private final Map<String, String> params;
+
+  /**
+   * Constructs a helper instance by parsing the provided reference ID.
+   *
+   * @param referenceId The string from {@code PredefinedContent#getReferenceID()}, expected to be
+   *     in URL query string format.
+   */
+  public PredefinedContentHelper(String referenceId) {
+    this.params = parse(referenceId);
+  }
+
+  /**
+   * Extracts and validates the 'event' value from the reference ID.
+   *
+   * @return An {@link Optional} containing the {@link DisplayNotificationEvent} if it is present
+   *     and valid, otherwise an empty Optional.
+   *     <pre>{@code
+   * PredefinedContentHelper helper = new PredefinedContentHelper("...&event=PIN_ENTERED");
+   * helper.getEvent().ifPresent(event -> System.out.println(event)); // Prints PIN_ENTERED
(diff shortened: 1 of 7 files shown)


src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    Map<String, String> queryPairs = new LinkedHashMap<>();
+    String[] pairs = referenceId.split("&");
+    for (String pair : pairs) {
+      int idx = pair.indexOf("=");
+      if (idx > 0 && idx < pair.length() - 1) {
+        String key = URLDecoder.decode(pair.substring(0, idx), StandardCharsets.UTF_8);
+        String value = URLDecoder.decode(pair.substring(idx + 1), StandardCharsets.UTF_8);
+        queryPairs.put(key, value);
```
thomasc-adyen: Is there any chance we have duplicate keys?

src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
```diff
@@ -0,0 +1,148 @@
...
+   * helper.getEvent().ifPresent(event -> System.out.println(event)); // Prints PIN_ENTERED
+   * }</pre>
+   */
+  public Optional<DisplayNotificationEvent> getEvent() {
+    return get(KEY_EVENT)
+        .flatMap(
+            eventValue -> {
+              try {
+                return Optional.of(DisplayNotificationEvent.valueOf(eventValue));
+              } catch (IllegalArgumentException e) {
```
thomasc-adyen, line 83: Should we log the exception  at least so it's not swallowed?

src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  /**
+   * Parses a URL query-like string into a map.
+   *
+   * @param referenceId The string to parse.
+   * @return A map of the parsed key-value pairs.
+   */
+  private static Map<String, String> parse(String referenceId) {
+    if (referenceId == null || referenceId.trim().isEmpty()) {
```
thomasc-adyen, line 135: Can we also log when this is null or empty?

src/main/java/com/adyen/util/tapi/SaleDataHelper.java
```diff
@@ -0,0 +1,52 @@
...
+   *     Optional if the field is absent or cannot be parsed.
+   */
+  public Optional<SaleToAcquirerData> getSaleToAcquirerData() {
+    String raw = saleData.getSaleToAcquirerData();
+    if (raw == null || raw.trim().isEmpty()) {
+      return Optional.empty();
+    }
+    try {
+      return Optional.of(SaleToAcquirerData.parse(raw));
+    } catch (Exception e) {
```
thomasc-adyen, line 49: Let's log the exception here

src/main/java/com/adyen/util/tapi/SaleToAcquirerData.java
```diff
@@ -0,0 +1,447 @@
...
+import java.util.logging.Logger;
+
+/**
+ * Sale information intended for the Acquirer, decoded from the Base64-encoded JSON string found in
+ * {@code SaleData.SaleToAcquirerData}.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class SaleToAcquirerData {
+
+  private static final ObjectMapper MAPPER =
```
thomasc-adyen: Do we need a dedicated object mapper here? Can we just use  `com.adyen.model.tapi.JSON`?

src/main/java/com/adyen/util/tapi/SaleToAcquirerData.java
```diff
@@ -0,0 +1,447 @@
...
+   *
+   * <ul>
+   *   <li>If the string is valid Base64 and decodes to a JSON object, it is parsed as JSON.
+   *   <li>Otherwise it is parsed as form-encoded key-value pairs.
+   * </ul>
+   *
+   * @param raw The raw {@code SaleToAcquirerData} string.
+   * @return Parsed {@link SaleToAcquirerData}.
+   */
+  public static SaleToAcquirerData parse(String raw) {
```
thomasc-adyen: I would extract this logic to its own class rather than have it be part of the DTO

src/main/java/com/adyen/util/tapi/PredefinedContentHelper.java
gcatanese: Added a fix and test

src/main/java/com/adyen/util/tapi/SaleToAcquirerData.java
gcatanese: Good call

src/main/java/com/adyen/util/tapi/SaleToAcquirerData.java
gcatanese: @thomasc-adyen This one seems better to leave as-is:

- `com.adyen.model.tapi.JSON` registers several features (like JavaTimeModule) that are not needed here
- `com.adyen.model.tapi.JSON` is auto-generated while the utility classes are not
- local `mapper` is light (set the only config it needs) and immutable (unlike tapi JSON)

jeandersonbc (approved):

PR #1963, opened 2026-05-15
[sessionauthentication] Code generation: update services and models

AdyenAutomationBot:
This PR contains the automated changes for the `sessionauthentication` service.

The commit history of this PR reflects the `adyen-openapi` commits that have been applied.


jeandersonbc (approved):
# [sessionauthentication] Model updates: resource type pruning and resilient enum handling

 ## Summary

 This update revises the `sessionauthentication` model package. Two concrete resource subtypes (`BalanceAccountResource`, `MerchantAccountResource`) are removed from the API surface, their corresponding enum values are dropped from `ResourceType`, and both `ProductType` and `ResourceType` switch from hard-failing on unknown enum values to logging a warning and returning `null`.

 ---

 ## Features Removed

 | Class / Enum | What was removed |
 |---|---|
 | `BalanceAccountResource` | Entire class deleted (`balanceAccountId` field, full builder/getter/setter API, JSON serialization helpers) |
 | `MerchantAccountResource` | Entire class deleted (`merchantAccountCode` field, full builder/getter/setter API, JSON serialization helpers) |
 | `ResourceType` | Enum values `BALANCEACCOUNT` (`"balanceAccount"`) and `MERCHANTACCOUNT` (`"merchantAccount"`) removed |
 | `Resource` | `@JsonSubTypes` entries and discriminator mappings for `balanceAccount` and `merchantAccount` removed |

 ---

 ## Behavioral Changes

 | Class | Before | After | Notes |
 |---|---|---|---|
 | `ProductType.fromValue()` | Throws `IllegalArgumentException` on unrecognised value | Logs a `WARNING` via `java.util.logging` and returns `null` | **Potential breaking change**  – callers that rely on the exception being thrown will silently receive `null` instead |
 | `ResourceType.fromValue()` | Throws `IllegalArgumentException` on unrecognised value | Logs a `WARNING` via `java.util.logging` and returns `null` | Same caveat as above |

 > **Breaking change notice:** Any code that catches `IllegalArgumentException` from `fromValue()` on  these enums, or that assumes a non-null return, will behave differently. NPEs can surface downstream if the returned `null` is used without a null-check.

PR #2025, opened 2026-08-07
Inject Pooling Connection Manager

danielrohe:
In a microservices especially based on Spring Boot it is often useful to inject a shared `PoolingHttpClientConnectionManager` into all clients including the AdyenHttpClient. The benefits are 

* users of the Adyen Java Library can configure the pool size correctly based on their needs on the shared pooling connection manager before injecting it into the AdyenHttpClient and do not need to rely on the quite low default configuration (max 25 connections to different Adyen API domains and max 5 connections to the same Adyen API domain). The default with max 5 connections at the same time to one Adyen API domain can cause latency issues because of pool exhaustion under high load
* especially within Spring Boot applications users of the Adyen Java Library can inject a pooling connection manager that is connected to telemetry systems like micrometer making it possible to observe the load on the connection pool used by the AdyenHttpClient.

Created the no-args constructor for backward compatibility and added one for injecting a shared `PoolingHttpClientConnectionManager`. Adjusted the `createHttpClientWithSocketFactory` method to consider the shared connetction pool manager if necessary.

### DIFF
--- a/src/main/java/com/adyen/httpclient/AdyenHttpClient.java
+++ b/src/main/java/com/adyen/httpclient/AdyenHttpClient.java
@@ -93,14 +93,29 @@ public class AdyenHttpClient implements ClientInterface {
 
   private static final String CHARSET = "UTF-8";
   private Proxy proxy;
-  private PoolingHttpClientConnectionManager sharedConnectionManager;
+  private final PoolingHttpClientConnectionManager sharedConnectionManager;
   private volatile CloseableHttpClient sharedHttpClient;
   private final Object lock = new Object();
 
   public AdyenHttpClient() {
-    super();
+    this(null);
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


gcatanese:
Thanks for the contribution @danielrohe . Injecting a connection manager is valuable for advanced use cases such as Micrometer telemetry, pool sharing, and lower-level Apache configuration.

Before merging, could you please:
- Fix the trailing whitespace reported by Checkstyle 
- Add tests proving that the injected manager is used, can be shared across clients, and remains open when an `AdyenHttpClient` is closed.
- Verify that custom pool limits remain unchanged and that the no-argument constructor retains its existing ownership and lifecycle behavior.

In a new issue we will expose in the `Config` object the pool-capacity configuration (`maxTotalConnections` and `maxConnectionsPerRoute`), I think those settings are also important in the default implementation offered by the library.

PR #2029, opened 2026-08-11
Make Swagger annotation dependencies optional

### LINKED ISSUE #2028: Make Swagger annotation dependencies optional
The SDK declares both Swagger v1 (`io.swagger:swagger-annotations`) and Swagger v3 (`io.swagger.core.v3:swagger-annotations`) as mandatory transitive dependencies. They annotate generated model metadata but are not required for API calls, serialization, webhook verification, or other SDK runtime behavior.

This causes avoidable conflicts for consumers that use a different Swagger namespace. A Spring Boot 4 application using springdoc Jakarta receives `swagger-annotations-jakarta` while Adyen adds the non-Jakarta v3 artifact containing the same annotation class names. In our packaged application this produced 62 duplicate-class collisions until both Adyen annotation artifacts were explicitly excluded.

The SDK compiles against the annotations, so marking them Maven-optional preserves SDK compilation and makes them available to consumers that explicitly want to reflect on Adyen model documentation. Ordinary consumers no longer receive two documentation-only artifacts.

Proposed change: add `<optional>true</optional>` to both Swagger annotation dependencies.

Verification performed:

* Full Adyen test suite on JDK 26: 617 tests, 0 failures/errors, 4 skips.
* Consumer Kotlin compilation with both dependencies excluded: Adyen payment/webhook and terminal modules compile successfully.

gregjotau:
Fixes #2028.

Mark the Swagger v1 and v3 annotation artifacts as optional. The SDK still compiles with both dependencies, while applications that only use Adyen APIs no longer receive documentation metadata transitively or collide with Jakarta Swagger annotations. Consumers that reflect on Adyen model annotations can declare the relevant artifact explicitly.

Verification:

* `mvn test -DskipITs` — 617 tests, 0 failures/errors, 4 skips on JDK 26.
* A Spring Boot 4/Kotlin consumer compiles its Adyen payment, webhook, and terminal modules with both Swagger dependencies excluded.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -307,11 +307,13 @@
             <groupId>io.swagger</groupId>
             <artifactId>swagger-annotations</artifactId>
             <version>${swagger-core-version}</version>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.swagger.core.v3</groupId>
             <artifactId>swagger-annotations</artifactId>
             <version>${swagger-annotations-v3-version}</version>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>jakarta.ws.rs</groupId>


gregjotau:
@jeandersonbc you can take over this. I dont' need ownership, I just want stuff done and efficient ;)

PR #2031, opened 2026-08-11
Avoid Gson final-field mutation on modern JDKs

### LINKED ISSUE #2030: Avoid Gson mutation of final ApplicationInfo field on modern JDKs
On JDK 26, `SaleToAcquirerDataSerializerTest` emits this warning while Gson deserializes terminal application info:

```text
WARNING: Final field adyenLibrary in class com.adyen.model.applicationinfo.ApplicationInfo has been mutated reflectively by class com.google.gson.internal.bind.ReflectiveTypeAdapterFactory...
WARNING: Mutating final fields will be blocked in a future release unless final field mutation is enabled
```

`ApplicationInfo` initializes `adyenLibrary` in its no-argument constructor, but Gson then populates the JSON value reflectively. The field being `final` therefore does not prevent mutation today; it only invokes deep reflection that modern JDKs are restricting (JEP 500): https://openjdk.org/jeps/500

This code is used by terminal `SaleToAcquirerData` serialization/deserialization. Applications should not need `--enable-final-field-mutation=ALL-UNNAMED` for ordinary SDK model handling.

Proposed fix: make this serialized field non-final, matching the rest of the Gson model. The public API remains unchanged (`getAdyenLibrary()` remains read-only).

Verification: the focused round-trip serialization test passes without the final-field warning, and the full 617-test suite passes on JDK 26.

gregjotau:
Fixes #2030.

Make the Gson-serialized `adyenLibrary` field non-final. Gson already mutates this field during deserialization; on JDK 26 that reflective write emits a final-field mutation warning and is scheduled to require explicit opt-in under JEP 500.

The public API remains read-only because no setter is added, and the constructor continues to initialize the SDK name/version.

Verification:

* `mvn -Dtest=SaleToAcquirerDataSerializerTest test` — round trip passes and the final-field warning is gone.
* `mvn test -DskipITs` — 617 tests, 0 failures/errors, 4 skips on JDK 26.

### DIFF
--- a/src/main/java/com/adyen/model/applicationinfo/ApplicationInfo.java
+++ b/src/main/java/com/adyen/model/applicationinfo/ApplicationInfo.java
@@ -28,7 +28,7 @@ import java.util.Objects;
 
 public class ApplicationInfo {
   @SerializedName("adyenLibrary")
-  private final CommonField adyenLibrary;
+  private CommonField adyenLibrary;
 
   @SerializedName("adyenPaymentSource")
   private CommonField adyenPaymentSource = null;


gcatanese:
Thank you @gregjotau for your contribution 🎉

gcatanese (approved):

PR #2033, opened 2026-08-11
Compile Java 11 artifacts with --release

### LINKED ISSUE #2032: Compile Java 11 artifacts with --release
The build currently configures `maven.compiler.source` and `maven.compiler.target` as 11. On a current JDK this produces the compiler warning:

> location of system modules is not set in conjunction with -source 11

`source` and `target` control language and class-file versions, but they do not restrict the JDK API symbol table. This leaves room for accidentally compiling against APIs introduced after Java 11 while still producing Java 11 bytecode.

Proposed solution:

- replace the source/target pair with `maven.compiler.release=11`
- retain the same Java 11 language and class-file baseline while also compiling against the Java 11 API surface

I verified this with `mvn clean test -DskipITs` on a current JDK: Maven reports `[debug release 11]`, the warning is gone, and all 617 tests pass (4 skipped).

Maven Compiler Plugin guidance: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

gregjotau:
Closes #2032.

Replace the separate Java source/target settings with `maven.compiler.release=11`. This preserves the Java 11 bytecode and language baseline while also preventing accidental linkage to JDK APIs introduced after Java 11.

Verification:

- `mvn clean test -DskipITs`
- compiler reports `[debug release 11]`
- the `system modules ... -source 11` warning is gone
- 617 tests pass, with 4 skipped

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,8 +24,7 @@
     </developers>
     <properties>
         <!-- Project Settings -->
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.release>11</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <sonar.organization>adyen</sonar.organization>


jeandersonbc (approved):
Hey @gregjotau, thank you not only for raising the issue but providing this PR (along with the Maven reference). We appreciate your time checking it! I also just verified myself for the warning (`mvn clean compile`) and using the `release` option. Changes look good to me!

gregjotau:
Great, we integrate Adyen into reai.no (accounting system) now and we want to contribute and improve all third party libraries we use. 

Especially important for us that dependencies are modularised in such a way that we bring in as little surface as possible :) 

thanks @jeandersonbc !

